PR #71, opened 2026-05-10
docs(rfc): import external chat history from Claude Code + Codex

chicoxyzzy:
## Summary

- New RFC at `docs/rfcs/import-external-chat-history.md` for one-shot ingest of Claude Code and Codex CLI session transcripts into Hecate's existing agent-chat store as read-only, searchable, attributable records.
- Indexed in `docs/rfcs/README.md`.

## Why

Operators run Claude Code and Codex outside Hecate and accumulate weeks of JSONL transcripts on disk (`~/.claude/projects/<slug>/<uuid>.jsonl`, `~/.codex/sessions/YYYY/MM/DD/rollout-*.jsonl`). Today there's no way to surface that history inside Hecate's Chats workspace, which means an operator who's adopted Hecate as the supervision surface for *new* agent work still has to flip back to `grep -r ~/.claude/projects` for *old* work.

## Shape of the proposal

- **One-shot, read-only.** No live mirroring, no editing, no resume — explicit non-goals listed in the RFC.
- **Same tables as live sessions.** Five new columns on `agent_chat_sessions` (`source_tool`, `source_file_path`, `source_file_size`, `source_file_mtime`, `imported_at`) + one partial unique index on `(source_tool, native_session_id)` constrained to imported rows. Live sessions are unaffected.
- **Two endpoints** under `/hecate/v1/agent-chat/imports/`: `scan` (read-only candidate listing) and `apply` (the actual write).
- **One UI entry point**: Settings → Chats → Import history modal. Imported sessions render in the existing transcript view with an "imported" badge and a read-only banner; the compose bar is hidden.
- **Source-format tables** for both tools, verified against actual files on disk.
- **Idempotent** via natural key + size/mtime change detection.

## Open questions called out in the RFC

- Workspace-slug ambiguity for Claude Code (paths containing `-`).
- System-prompt storage (Claude Code's are ~100KB; v1 verbatim, v2 hash-and-dedupe-table).
- PII / secret scrubbing posture (lean: don't scrub, document in `docs/security.md`).
- Channel taxonomy for Codex's `developer`-role messages (skill / plugin injections).
- CLI vs UI entry point (UI in v1; CLI is a v2 add).

## Status

Design notes only. Not implemented. The RFC is the artifact — no code changes, no schema migration applied.

## Test plan

- [ ] Reviewers read `docs/rfcs/import-external-chat-history.md` end-to-end.
- [ ] Reviewers sanity-check the source-format tables against their own `~/.claude/projects/` and `~/.codex/sessions/` files (versions drift; the RFC promises a permissive parser, but the table should still be roughly accurate).
- [ ] Reviewers raise objections to the non-goals — particularly "no resume of imported sessions," which is the most likely surface to push back on.


chicoxyzzy:
@copilot review the squashed commit. Three rounds of corrections folded into one. Highlights vs the original draft:

- Glob fixed: `~/.claude/projects/*/*.jsonl` (per-workspace subdirectories).
- Schema additions follow Hecate's existing additive `ensureSessionColumn` pattern, not standalone SQL files. Time columns are `TIMESTAMP`, not `INTEGER`.
- Partial-unique-index name derives from the (possibly prefixed) `sessionsTable` like `messagesIndex` / `sessionsIndex` do — no hard-coded literal.
- Activities stay JSON on `agent_chat_messages`; no `agent_chat_activities` table is introduced.
- System prompts map to a leading `Role: "system"` message — `Session` struct gets no new `SystemPrompt` field.
- API accepts paths/directories only (no blob upload).
- Rollback via restore-from-snapshot per `migration-cli.md` (additive-only migrations, no down-migrations).
- `docs/security.md` is a markdown link, not inline code.

PR #72, opened 2026-05-10
docs(rfc): observability workspace redesign

chicoxyzzy:
## Summary

- New RFC at `docs/rfcs/observability-redesign.md` proposing a two-tab IA for the Observability workspace: **Live** ("is the system OK right now?") and **Inspect** ("what happened with this request?").
- Cross-surface drill-down: chats, agent-chat runs, and task runs link to their producing trace; the trace drawer links back.
- Operator-actionable filters under Inspect (errors-only, time range, route reason, attribute key=value, request-id substring) plus a phase-rollup viz across the filtered set.
- One small backend addition for Live (`/hecate/v1/system/runtime/timeline` for error-rate + latency sparklines); everything else is frontend on existing data.
- Indexed in `docs/rfcs/README.md`.

## Why

Today's Observability view is a single stack — stat strip + recent-traces table + drawer — optimized for the question "show me a recent request, let me poke at it." Real operator questions don't map cleanly:

- "Is the system OK right now?" → no error-rate / latency trend, no isolated failed-traces feed.
- "Why did this chat go wrong?" → chats carry `trace_id` but no UI links from a chat to its trace.
- "Where is time going across all my recent calls?" → no aggregation across the visible set; you click 50 waterfalls.
- "Just show me failures." → no errors-only filter.

The drawer also stacks five panels of unequal importance with nothing telling the operator which to read first.

## Tier 1 prerequisite (separate PR)

The 1214-line `ObservabilityView.tsx` and the span-rendering math in `ui/src/lib/runtime-trace.ts` need a cleanup pass before this redesign lands. Concrete bugs that the cleanup closes:

- `t0 = Math.min(...starts)` collapses to 0 when any span has unparseable `start_time`, blowing the entire waterfall to the right (single bad span breaks the whole render).
- "Critical path" is actually "longest single descent" — wrong with sibling parallelism.
- Span sort by `(startMs, depth)` orphans children visually when clock skew puts a child before its parent.
- Negative-duration spans (clock skew) clamp to 1ms instead of being marked "?".
- Trace-detail fetch retries every 2s with no cap.
- Multiple roots silently drop all but the first from critical-path highlighting.

This RFC starts where Tier 1 ends. Tier 1 is independent and reverts cleanly.

## Phasing in this RFC

1. Two-tab IA (Live + Inspect; existing content under Inspect).
2. Filter bar: time range, errors-only, status, route reason, request-id substring.
3. Cross-surface drill-down (`<TraceLink>` component).
4. Phase rollup viz under Inspect.
5. Live tab content + new `/system/runtime/timeline` endpoint.
6. Attribute key=value filter (visible-set autocomplete first; "scan deeper" follows).

Each phase is a real PR.

## Open questions called out in the RFC

- Live → Inspect transition: tab switch vs modal-on-Live.
- Distinguishing "no data because nothing happened" from "no data because retention pruned it."
- Phase-rollup math: linear vs log axis (provider phase dominates by 10x).
- Error-rate bucket size: fixed bucket count vs fixed bucket duration.
- Attribute autocomplete scope: client-side over visible set vs server-side over retention window.
- Mobile / narrow-viewport behavior for the Live tab's three-row layout.

## Status

Design notes only. Not implemented. The RFC is the artifact — no code changes, no schema migration, no endpoint added by this PR.

## Test plan

- [ ] Reviewers read `docs/rfcs/observability-redesign.md` end-to-end.
- [ ] Reviewers push back on the two-tab IA: is the Live/Inspect split worth the navigation cost vs an enriched single view?
- [ ] Reviewers sanity-check the Tier 1 bug list against their own experience with the current waterfall.
- [ ] Reviewers raise objections to the non-goals — particularly "no full metrics dashboard" (some operators may want this) and "no alerting" (some may consider it table-stakes).


chicoxyzzy:
Closing in favor of shipping directly. Design review reduced the scope significantly — the actually-missing pieces are smaller than the RFC proposed (the Status dropdown already does errors-only filtering, the recent-failures feed is just that filter applied), and what's left fits in two small frontend PRs:

1. `<TraceLink>` component + chat→trace drill-down (the highest-leverage missing piece).
2. Two sparklines above the trace table (error rate + latency, derived from the existing recent-traces feed, no backend change).

The two-tab IA, the new `/system/stats/timeline` endpoint, the phase rollup viz, and the attribute autocomplete — all over-scoped for the actual problem. If any of them become a real ask later, that's a separate RFC.

Working Rule 2: simplicity first.

PR #77, opened 2026-05-10
feat(desktop): wire Tauri auto-update — auto-check + dismissible banner

chicoxyzzy:
## Summary

End-to-end auto-update for the desktop app: silent check on launch, dismissible banner if a newer release is found, click installs and restarts. Gated on a maintainer keypair you generate once and a one-line follow-up commit to flip `active: true` after the pubkey lands.

Today the updater plugin is loaded in Rust but `active: false` with empty endpoint and pubkey. This PR wires every layer so once the keypair is in place, the next tagged release automatically signs the bundles and emits `latest.json`.

## What's in this PR

**Frontend (gateway UI, runs inside the webview):**
- `useDesktopUpdate()` hook in `ui/src/lib/desktop-update.ts` — fires once on mount in the Tauri runtime via dynamic `import()`, otherwise no-op. Errors swallowed (no pubkey, network down, signature mismatch all collapse to "no update").
- `<UpdateBanner />` in `ui/src/features/shared/UpdateBanner.tsx` — top-of-app banner with **Install and Restart** / **Dismiss**. Dismiss persists for the session via `sessionStorage`. Shows live download progress (text + thin bar) while installing.
- Wired into `AppShell.tsx` above the existing error banner. Reuses the existing `.page-banner` CSS pattern with a new `--update` modifier.
- `isTauriRuntime()` lives in `ui/src/lib/tauri.ts` (shared with `App.tsx`).

**Tauri config:**
- `tauri.conf.json`: endpoints set to `https://github.com/hecatehq/hecate/releases/latest/download/latest.json`. `pubkey` stays empty and `active` stays `false` — the one-line follow-up commit (after you generate the key) flips both.
- `capabilities/default.json`: adds `updater:default` so the frontend can call updater APIs.

**CI:**
- `.github/workflows/_tauri-shared.yml`: passes `TAURI_UPDATER_PRIVATE_KEY` + `TAURI_UPDATER_PRIVATE_KEY_PASSWORD` to tauri-action for release-workflow runs only. Same caller-side / called-side protection model as the Apple secrets — PR validation builds never see them. With the secrets present, tauri-action signs each platform bundle and produces `latest.json` as a release asset alongside the `.dmg` / `.deb` / `.AppImage` / `.msi`.

**Docs:**
- `docs/desktop-updater-signing.md` — new maintainer-side checklist (parallel to `docs/macos-signing.md`): generate keypair, add secrets, paste pubkey, flip `active`. Includes verification recipe and rotation playbook.
- `docs/desktop-app.md` — updated the "what doesn't yet" bullet to describe the wired-but-inactive state; dropped the matching Tier-2 roadmap row.

## What you need to do once

1. `bunx @tauri-apps/cli signer generate -w ~/.tauri/hecate-updater.key` — pick a strong password, save it.
2. Add to repo secrets:
   - `TAURI_UPDATER_PRIVATE_KEY` = full contents of the `.key` file
   - `TAURI_UPDATER_PRIVATE_KEY_PASSWORD` = the password from step 1
3. One-line PR: paste the contents of `~/.tauri/hecate-updater.key.pub` into `tauri.conf.json`'s `updater.pubkey`, flip `active` to `true`.

Once that lands and the next release tag is cut, every existing install (built with the same pubkey embedded) sees a banner on next launch.

## Verification

- `bun run test`: 683 passing (+9 new in `ui/src/lib/desktop-update.test.ts` covering no-check-outside-Tauri, dismiss flow, sessionStorage skip, error swallowing, install-and-restart toggling, progress events, and an explicit StrictMode guard test).
- `bun run build`: clean. Index chunk +1.4 KB (the dynamic-import wrapper) + the progress UI.
- The dynamic import keeps the plugin out of the web bundle path; non-Tauri runtimes get a no-op hook.

## Test plan

- [x] Unit + build clean.
- [ ] After you do the one-time setup and cut a release: install the previous version on a clean machine, launch it, confirm the banner appears within ~5 seconds and clicking **Install and Restart** completes the upgrade with visible progress.
- [ ] Confirm dismissed banner stays dismissed within a session and reappears after a fresh launch (sessionStorage scope).
- [ ] Confirm web build (browser tab pointing at gateway) and Docker run never show the banner — gated on `isTauriRuntime()`.

## Notes / decisions

- **Endpoint**: GitHub Releases asset, not a custom redirect — zero new infrastructure, secret-free fetch from the client.
- **Cadence**: silent check on launch, banner if found. No menu item, no auto-install, no forced restart. Dismissible per-session.
- **Channel**: single channel for now (`latest`). If alpha cadence becomes too noisy, the next iteration can introduce a stable / preview switch — out of scope here.
- **Install-on-quit**: deliberately not implemented. The Tauri 2 updater plugin only exposes immediate install + relaunch on every platform; deferred install would need custom per-platform staging code (~2-3 days). Out of scope here. The button label is "Install and Restart" — explicit about the trade.
- **`active: false` ships in this PR by design**: lets the wiring land and be reviewed without forcing keypair generation in the same PR. Flipping it is a one-line follow-up after secrets are in place.

### DIFF
--- a/.github/workflows/_tauri-shared.yml
+++ b/.github/workflows/_tauri-shared.yml
@@ -194,6 +194,17 @@ jobs:
           APPLE_PASSWORD: ${{ matrix.os == 'macos-latest' && inputs.tagName != '' && secrets.APPLE_PASSWORD || '' }}
           APPLE_TEAM_ID: ${{ matrix.os == 'macos-latest' && inputs.tagName != '' && secrets.APPLE_TEAM_ID || '' }}
           KEYCHAIN_PASSWORD: ${{ matrix.os == 'macos-latest' && inputs.tagName != '' && secrets.KEYCHAIN_PASSWORD || '' }}
+          # Updater signing key. Same caller-side / called-side
+          # protection model as the Apple secrets above: scope to a
+          # release-workflow run with a non-empty tagName. When set,
+          # tauri-action signs the platform bundles and emits a
+          # latest.json manifest as a release asset that the in-app
+          # updater consumes. When empty (PR validation, missing
+          # secrets), unsigned bundles ship and the updater stays
+          # inert. See docs/desktop-updater-signing.md for the
+          # keypair custody checklist.
+          TAURI_SIGNING_PRIVATE_KEY: ${{ inputs.tagName != '' && secrets.TAURI_UPDATER_PRIVATE_KEY || '' }}
+          TAURI_SIGNING_PRIVATE_KEY_PASSWORD: ${{ inputs.tagName != '' && secrets.TAURI_UPDATER_PRIVATE_KEY_PASSWORD || '' }}
         with:
           projectPath: tauri
           tagName: ${{ inputs.tagName }}
--- a/docs/desktop-app.md
+++ b/docs/desktop-app.md
@@ -69,8 +69,14 @@ What doesn't yet:
 - No Homebrew formula or cask yet. A formula would help CLI installation, and
   a cask would help app distribution. macOS now signs+notarizes via
   `APPLE_*` repo secrets; a cask would still be additional polish.
-- No auto-update — plugin is wired but `active: false` until an update
-  endpoint is decided.
+- Auto-update is wired end-to-end (endpoint, capability, frontend
+  banner, signing in CI) but `active: false` until the operator
+  generates the updater keypair and pastes the pubkey into
+  `tauri.conf.json`. See [`desktop-updater-signing.md`](desktop-updater-signing.md)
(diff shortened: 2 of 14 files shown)


chicoxyzzy:
@copilot review the latest commit (review: address PR #77 comments — dedupe, guard, tests). Three concerns from the previous round addressed: `isTauriRuntime()` extracted to `ui/src/lib/tauri.ts`; StrictMode guard via `useRef`; new `ui/src/lib/desktop-update.test.ts` with 9 cases including an explicit StrictMode-double-invoke test.

PR #79, opened 2026-05-10
feat(agent-adapters): guide Claude Code ACP credential setup

chicoxyzzy:
## Summary

Adds a guided Claude Code ACP credential setup path and folds the earlier auth-readiness clarification into one commit.

- Treat Claude Code config-file presence as `auth_status=unknown` because it can make the standalone `claude` CLI work while `claude-agent-acp` still fails with `Authentication required`.
- Add encrypted local adapter credentials to the control-plane store, persisted by both memory and SQLite backends.
- Add allowlisted adapter credential endpoints and expose only `credential_configured` plus a redacted preview in adapter status rows.
- Inject stored adapter credentials into both Settings probes and real ACP sessions after env sanitization.
- Add a Settings -> External agents guided setup card for Claude Code: copy `claude setup-token`, paste token, save, then re-test.
- Keep Claude billing/subscription probe failures from being rewritten as auth-token setup guidance.
- Update runtime API and external-agent docs for the guided setup and credential endpoints.

## Verification

- `go vet ./...`
- `go test ./...`
- `cd ui && bun run test`

### DIFF
--- a/docs/external-agent-adapters.md
+++ b/docs/external-agent-adapters.md
@@ -36,7 +36,7 @@ own the `hecate-acp` bridge process.
 | Adapter | How Hecate starts it | Auth expected by the underlying agent |
 |---|---|---|
 | Codex | Hecate-managed launcher for `@zed-industries/codex-acp` via local `npx`; direct `codex-acp` also works | Codex CLI / adapter login or config |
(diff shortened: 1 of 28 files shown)


chicoxyzzy:
@copilot review

Please re-review after the squash/force-push. In particular, the Claude-specific probe hint is now gated so billing/subscription errors keep their original diagnostic instead of being rewritten as auth-token setup guidance.

PR #80, opened 2026-05-10
fix(ci): stitch latest.json across matrix legs so auto-update works

chicoxyzzy:
## Summary

Fixes the missing `latest.json` from `v0.1.0-alpha.24` — the first release with the updater wired, but no manifest got published because `tauri-action`'s `includeUpdaterJson` mode can't build a complete manifest from a single matrix leg.

## What was broken

The macOS leg's log on the alpha.24 release:

```
Finished 1 bundle at:
/.../Hecate.app.tar.gz
/.../Hecate.app.tar.gz.sig
Found artifacts:
/.../Hecate_0.1.0-alpha.24_aarch64.dmg
Found release with tag v0.1.0-alpha.24.
Uploading Hecate_0.1.0-alpha.24_aarch64.dmg...
Signature not found for the updater JSON. Skipping upload...
```

Each leg has its own platform's `.sig` but not the others'. `tauri-action` correctly refuses to ship a partial manifest (a `latest.json` missing entries would break auto-update on the missing platforms). But that means **no `latest.json` at all** in matrix mode.

On top of that, tauri-action's release-upload step only handles the platform bundles (`.dmg` / `.deb` / `.AppImage` / `.msi`) — not the updater payloads (`.app.tar.gz` / `.AppImage.tar.gz` / `.msi.zip`). So even if `latest.json` existed and pointed at the right URLs, the URLs would 404.

Result on alpha.24's release page: 4 platform bundles, 0 updater files, 0 manifest.

## What changes

Two additions to `.github/workflows/_tauri-shared.yml`:

1. **Per-leg upload of updater payloads.** After `tauri-action` runs, glob `**/*.tar.gz`, `**/*.zip`, `**/*.sig` in the bundle output and `gh release upload --clobber` them. Gated on `inputs.tagName != ''` so PR validation is unaffected.

2. **New `publish-updater-manifest` job** runs after the matrix (`needs: build`). Downloads every `*.sig` from the release, reads each, constructs `latest.json` with platform → `{signature, url}` for `darwin-aarch64` / `linux-x86_64` / `windows-x86_64`, uploads via `gh release upload`. Exits 1 if any platform's signature is missing — better to fail loudly than ship a manifest that silently breaks one platform's auto-update.

## Effect

- **From the next signed release**, the GitHub Release page will include `Hecate.app.tar.gz` + `.sig`, `Hecate_X.Y.Z_amd64.AppImage.tar.gz` + `.sig`, `Hecate_X.Y.Z_x64_en-US.msi.zip` + `.sig`, **and** `latest.json` with all three signatures and URLs.
- **`v0.1.0-alpha.24` users won't see the banner** until they manually upgrade once to the first release with this fix. After that, auto-update works for every subsequent release.

## Test plan

- [x] `python3 -c "import yaml; yaml.safe_load(open(...))"` confirms the YAML parses cleanly (used Ruby's `yaml` since python3-yaml wasn't installed; same result).
- [ ] Cut a real signed release (`v0.1.0-alpha.25`) and confirm:
  - All three `.sig` + payload pairs appear as release assets.
  - `latest.json` appears with the right shape.
  - The shape passes Tauri's updater signature verification (operator's existing alpha.24 install, manually upgraded to alpha.25, then sees the alpha.26 banner).
- [ ] PR-validation run on this PR confirms the `publish-updater-manifest` job is skipped when `tagName` is empty (existing tauri-build.yml passes `tagName: ''`).

### DIFF
--- a/.github/workflows/_tauri-shared.yml
+++ b/.github/workflows/_tauri-shared.yml
@@ -210,6 +210,52 @@ jobs:
           tagName: ${{ inputs.tagName }}
           args: --bundles ${{ matrix.bundles }}
 
+      # Upload updater payloads (.tar.gz / .zip + .sig) to the GitHub
+      # Release alongside the platform bundles. tauri-action's default
+      # release-upload step only handles the explicit platform bundles
+      # (.dmg/.deb/.AppImage/.msi); it doesn't upload `.sig` files or
+      # the macOS `.app.tar.gz` updater payload that tauri's bundler
+      # produces when an updater signing key is configured. Without
+      # this step:
+      #   - macOS: latest.json's `.app.tar.gz` URL would 404.
+      #   - Linux/Windows: the raw `.AppImage`/`.msi` are uploaded by
+      #     tauri-action, but their matching `.sig` files aren't —
+      #     latest.json's signature field would point at nothing.
+      # We over-upload (also grab wrapped `.AppImage.tar.gz` /
+      # `.msi.zip` and their `.deb`/`.rpm` siblings) because filtering
+      # to exactly the referenced files would be brittle across tauri
+      # bundler versions. The stitch job picks the canonical raw forms.
+      # Skipped on PR-validation runs (no tagName).
+      #
+      # The macOS Hecate.app.tar.gz has no version in its filename — each
+      # tag's release has its own copy, so versions don't clobber each
+      # other across releases. The --clobber flag below only matters for
+      # re-running this workflow on the same tag (rare; lets us re-upload
+      # a corrected build without manual cleanup).
+      #
+      # macOS-latest ships bash 3.2, which lacks `shopt -s globstar` and
+      # `mapfile`. We use `find` + positional-param accumulation to stay
+      # 3.2-compatible. Process substitution `< <(...)` IS in 3.2.
+      - name: Upload updater payloads to release
+        if: inputs.tagName != ''
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          cd tauri/src-tauri/target/release/bundle
+          set --
+          while IFS= read -r f; do
+            set -- "$@" "$f"
+          done < <(find . \( -name '*.tar.gz' -o -name '*.zip' -o -name '*.sig' \) -type f | sort)
+          if [ $# -eq 0 ]; then
+            echo "no updater payloads produced for ${{ matrix.target }} — skip upload"
+            exit 0
+          fi
+          printf '  - %s\n' "$@"
+          gh release upload "${{ inputs.tagName }}" "$@" \
+            --clobber \
+            --repo "${{ github.repository }}"
+
       # Persist as workflow artifact for PR runs so reviewers can download
       # and test-launch from the run page. Skipped on tagged releases —
       # those upload directly to the GitHub Release.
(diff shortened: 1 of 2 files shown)


chicoxyzzy:
@copilot review the follow-up commit. Four round-1 concerns addressed:

- **bash 3.2 on macOS:** rewrote the per-leg "Upload updater payloads" step without `shopt -s globstar` or `mapfile`. Uses `find` + positional-param accumulation (`set -- "$@" "$f"`) so it works on macOS-latest's default bash 3.2.
- **CR in .sig files:** `tr -d '\r' < "$sig"` strips CRs that Windows runners can leave in. Bash's `$(...)` only strips trailing newlines, not CRs — a stray CR in the JSON `signature` string breaks minisign verification.
- **for-loop on missing glob:** added `shopt -s nullglob` in the stitch step before iterating `sigs/*.sig`. Empty input now skips the loop cleanly and the explicit "missing updater signature(s)" check surfaces a more actionable error.
- **Misleading comment:** reworded the "clobbers across releases" line to say "--clobber matters only for re-runs of the same tag's workflow" — release assets are scoped per-tag, so different tags don't overwrite each other's identically-named assets.

chicoxyzzy:
Round 2: reconciled Codex's review against the Tauri 2 plugin-updater source.

**Codex flagged two things**:
1. `bundle.createUpdaterArtifacts: true` is missing from `tauri.conf.json`.
2. `latest.json` shouldn't point at `.AppImage.tar.gz` / `.msi.zip` — Tauri 2 canonical form is raw `.AppImage` / `.msi`.

**Resolution**:
1. *Not adopted.* alpha.24's CI logs show the bundler emits `.sig` files for both raw and wrapped forms without that flag being set — so the empirical fix doesn't depend on it. Adding it would be a behaviour-change in a PR that's supposed to be ops-layer-only.
2. *Adopted.* Switched `latest.json` URLs to raw `.AppImage` / `.msi` (see `aecbcc34`). The wrapped forms still ship as part of the over-upload (filtering them out would be brittle across bundler versions), but they're no longer referenced. macOS stays on `.app.tar.gz` because `install_inner()` in `plugins/updater/src/updater.rs` unconditionally does GzDecoder + tar::Archive — there is no raw form to target on macOS.

Source-of-truth pointers (Tauri 2 updater plugin):
- macOS `install_inner` — `.app.tar.gz` only
- Linux `install_appimage` — magic-byte detection, accepts both raw and wrapped
- Windows `extract`/`install_msi` — same magic-byte detection

Please re-review `.github/workflows/_tauri-shared.yml`, in particular the case statement in the `Build latest.json` step and the rationale comment above it. The expected failure modes to consider:
- Future bundler version stops emitting wrapped forms → still fine (we now point at raw).
- macOS leg fails to produce `.app.tar.gz.sig` → caught by the explicit `missing` check; manifest publish fails loudly rather than shipping a broken one.
- New unexpected payload type appears in the bundle dir → falls into the `*) echo "unexpected …" ;;` branch and is visible in the log without failing the build.

@copilot please re-review.

chicoxyzzy:
Round-2 update + reconciliation with Codex's follow-up:

**Empirical check.** Pulled the alpha.24 release-build logs. The bundler **did** produce `.sig` files on every platform:

```
macOS:   Hecate.app.tar.gz.sig
Linux:   *_amd64.AppImage.sig, *_amd64.AppImage.tar.gz.sig, *_amd64.deb.sig, *.rpm.sig
Windows: *_x64_en-US.msi.sig, *_x64_en-US.msi.zip.sig, *_x64-setup.exe.sig, *_x64-setup.nsis.zip.sig
```

Followed immediately by tauri-action's `Signature not found for the updater JSON. Skipping upload...` — i.e., the `.sig` files exist on disk after the build, they just never get uploaded to the release because tauri-action's per-leg manifest builder can't see the other platforms' sigs from a single matrix leg. That's the upload gap PR #80 closes. The release having no `.sig` files (which is true and verifiable from the asset list) is consistent with this — not with "bundler didn't produce them."

**But Codex's recommendation is still right** for a different reason. The bundler produces v1-compatible artifacts today because of some default (tauri-action's, tauri-cli's, or a transitive setting) that's not documented in our config. That's an implicit dependency we shouldn't rely on. New commit (`6...` — pushed) adds it explicitly:

```json
"bundle": {
  "createUpdaterArtifacts": "v1Compatible",
  …
}
```

**Why `"v1Compatible"` and not `true`** (this is load-bearing — the stitch job depends on it):

- `true` produces v2-native wrapped artifacts only: `.AppImage.tar.gz` / `.msi.zip` (same shape as macOS `.app.tar.gz`). Under that mode the raw `.AppImage` / `.msi` do **not** get a `.sig` sibling.
- `"v1Compatible"` produces v2-native artifacts AND signs the raw platform bundles directly. Both forms get a sig.
- The stitch job's round-2 change points `latest.json` at raw `.AppImage` / `.msi` — those URLs only have working signatures under `"v1Compatible"`.

If Codex's suggestion of `true` had been adopted as written, the manifest would point at `.AppImage` and `.msi` URLs whose `.sig` files weren't produced. That's why I'm pushing back on the exact value while accepting the broader concern.

**Verification plan.** Next signed release should now show in the assets:
- `Hecate.app.tar.gz` + `.sig`
- `*.AppImage.sig` (next to the existing `*.AppImage`)
- `*.msi.sig` (next to the existing `*.msi`)
- `latest.json`

@copilot please re-review with both the workflow change and the `tauri.conf.json` change in scope.

chicoxyzzy:
**Correction: my previous comment misread the build log.**

I cited paths under tauri-action's \`Looking for artifacts in:\` section as proof the bundler produced \`.sig\` files. That section enumerates **candidate paths** to scan, not files that exist. The authoritative section is \`Found artifacts:\` and on alpha.24 it contained only the installers — zero \`.sig\` files on any leg:

\`\`\`
macOS:   Hecate_0.1.0-alpha.24_aarch64.dmg            (only)
Linux:   Hecate_0.1.0-alpha.24_amd64.deb + .AppImage  (only)
Windows: Hecate_0.1.0-alpha.24_x64_en-US.msi          (only)
\`\`\`

Codex (and the reviewer relaying it) were right.

**Why no sigs despite \`TAURI_SIGNING_PRIVATE_KEY\` being set:** the env var was injected (log shows \`TAURI_SIGNING_PRIVATE_KEY: ***\`), but with \`createUpdaterArtifacts\` at its default \`false\`, the bundler produces no updater artifacts, so \`sign_updaters\` has nothing to sign and early-returns. The signing key was present but inert. This is exactly what Codex described.

**That makes \`f1cec361\` (createUpdaterArtifacts = "v1Compatible") the load-bearing fix in this PR**, not a defensive add-on. Without it, the workflow upload step has nothing to upload and the stitch job fails on the explicit \`missing updater signature(s)\` check — better than alpha.24's silent failure, but still no auto-update.

**What still stands from my prior comment:**
- The choice of \`"v1Compatible"\` over \`true\` is still load-bearing. The stitch job points \`latest.json\` at raw \`.AppImage\` / \`.msi\`; \`true\` only signs wrapped artifacts. \`"v1Compatible"\` signs both raw and wrapped.
- Raw URL mapping in the stitch job is still correct.
- The upload step is still needed (tauri-action doesn't upload \`.sig\` files even when they exist).

**Confirmed verification plan for the next signed release** (no change from prior comment, but now grounded in the right diagnosis): release assets should gain \`Hecate.app.tar.gz\` + \`.sig\`, \`*.AppImage.sig\`, \`*.msi.sig\`, and \`latest.json\`. If any are missing, the stitch job fails loudly instead of shipping a broken manifest.

@copilot — the PR contents haven't changed since the last re-review request, this is a correction to my comment thread. The three commits still in scope:
- \`74604f11\` — bash 3.2 / nullglob / CR / comment
- \`aecbcc34\` — raw \`.AppImage\` / \`.msi\` URLs
- \`f1cec361\` — explicit \`createUpdaterArtifacts: "v1Compatible"\` (load-bearing)

PR #81, opened 2026-05-11
fix(ui): collapse same-request_id traces in Observability table + drawer polish

chicoxyzzy:
## Summary

In Observability, the Recent Requests table was rendering one row per *trace*, but a single user-facing request can produce multiple internal traces all sharing the same \`request_id\`. In practice a chat send produced four or five visually-identical rows (provider/model both \"—\", same truncated ID). Clicking any of them opened the same detail panel because \`traces.find()\` returned the first array match regardless of which row was clicked.

This PR collapses rows by \`request_id\` and folds in a handful of related polish items from the same audit pass:

- **Group by request_id.** One row per request. The chosen representative is the trace with the most spans (tie-break: errors > ok > unset). Collapsed siblings surface as a \`+N\` badge next to the request ID.
- **Drawer mirrors the row.** \`selectedTrace\` resolves through \`groupedTraces\` first, so the panel's latency and status match the row instead of whichever sibling happened to be first in the array.
- **Friendlier drawer header.** Was \`DEE151F1… · —/—\` when no provider was selected. Now reads \`no provider selected (tried <X>)\` if there's a rejected candidate, or \`request\` if there's no route context at all.
- **Rejected-candidate hints in the table.** When the route skipped every candidate, Provider/Model show the rejected candidate name muted with a tooltip explaining the skip reason instead of a bare \`—\`.
- **★ critical-path legend.** The star marker had only a hover tooltip; added an explicit \`★ critical path\` chip to the phase legend whenever any span is on the critical path.
- **Waterfall padding alignment.** The sticky ruler's \`1fr\` column was 4px wider than each SpanRow's \`1fr\` column (rows had \`paddingRight: 4\`, ruler had none). Time-axis ticks now line up with the bar edges.

Pinned by a new \`ObservabilityView\` test that asserts five traces sharing one \`request_id\` collapse to a single row with the highest-span-count representative and a \`+4\` badge.

## Test plan

- [x] \`bun run test\` — all 687 UI tests pass; one new test added.
- [x] Manual verification against a live dev server with a chat session that produced five traces under one request_id. Confirmed:
  - Table shows one row (was 5).
  - \`+4\` badge appears next to the request ID.
  - Drawer header shows the request label (not \`—/—\`).
  - Drawer latency matches the row's latency.
  - \`★ critical path\` chip visible in the Spans legend.
  - Rightmost time tick aligns with bar edges in the waterfall.
- [ ] Reviewer to sanity-check the dedup heuristic (most spans + status_code tie-break) against any real trace traffic where the \"main\" call might *not* be the one with the most spans.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/ui/src/features/overview/ObservabilityView.test.tsx
+++ b/ui/src/features/overview/ObservabilityView.test.tsx
@@ -217,6 +217,45 @@ describe("ObservabilityView", () => {
     expect(container.textContent).toMatch(/\+4/);
   });
 
+  it("never renders 'undefined' when route candidates are partially populated", async () => {
+    // Defensive against route.candidates entries with missing provider
+    // or model fields — both are optional on the runtime type. A
+    // half-populated candidate must NOT reach the rendered output as
+    // literal "undefined" in the table cells or drawer header.
+    fetchMock.mockImplementation(tracesFetchHandler(
+      [{
+        request_id: "req-partial-cand",
+        started_at: new Date().toISOString(),
+        span_count: 1,
+        status_code: "ok",
+        route: { candidates: [{ outcome: "skipped", skip_reason: "unsupported_model" }] },
+      }],
+      {
+        request_id: "req-partial-cand",
+        started_at: new Date().toISOString(),
+        spans: [],
+        route: { candidates: [{ outcome: "skipped", skip_reason: "unsupported_model" }] },
+      },
+    ));
+    const state = createRuntimeConsoleFixture({ session: localSession });
+    let container = null as unknown as HTMLElement;
+    await act(async () => {
+      const result = render(<ObservabilityView state={state} actions={createRuntimeConsoleActions()} />);
+      container = result.container;
+    });
+    await waitFor(() => {
+      expect(container.textContent).toMatch(/req-part/);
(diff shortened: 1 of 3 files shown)


chicoxyzzy:
Round 2 — addressed all three Copilot comments in `6c…`:

- **`RecentActivityStrip` dedup**: now consumes `groupedTraces.map(g => g.entry)` instead of `filteredTraces`. The strip and the table show the same deduped set, so dot keys no longer collide on shared `request_id`.
- **Drawer title `tried undefined`**: added `rejected.provider || "—"` fallback.
- **Provider/Model cell undefined-leaks**: tightened the candidate selector to `outcome === "skipped" && !!c.provider`, and gated the Model cell on `rejected?.model` so a missing-model entry falls through to "—" instead of rendering blank.

New regression test pins the behavior: a trace with a partially-populated route candidate (no `provider`/`model` fields) must not produce the literal "undefined" anywhere in the rendered output, both in the table cell and in the drawer header after row click.

`bun run typecheck` clean, `bun run test` 689/689 passing.

@copilot please re-review.

PR #84, opened 2026-05-11
fix(ui): polish observability trace and runtime UX

chicoxyzzy:
## Summary

- Rework the trace drawer waterfall into a DevTools-style timeline with aligned ticks, grid lines, sticky local axis, real sub-ms scaling, and request bars that use the full available width.
- Make Recent requests easier to read: provider/model are recovered from route spans or request ledger when route fields are missing, request IDs are compact in the table but copy the full value, latency is request-level across grouped sibling traces, and Reason/Fallback are collapsed into one compact Route column.
- Simplify Observability filters: provider stays explicit, model now starts at All models and filters observed traffic instead of implying the current chat model is the scope; Live/Paused no longer shifts the header.
- Clarify runtime cards with operator language, grouped sections, helper tooltips, status labels like Idle/Disabled/Memory store, and a short task-runtime summary.
- Remove confusing waterfall affordances: critical-path star, parent-span outline style, and mispositioned global timeline scrolling.

## Verification

- `cd ui && bun run test`

### DIFF
--- a/ui/src/features/overview/ObservabilityView.test.tsx
+++ b/ui/src/features/overview/ObservabilityView.test.tsx
@@ -200,6 +200,37 @@ describe("ObservabilityView", () => {
     // Provider names appear
     expect(container.textContent).toMatch(/openai/);
     expect(container.textContent).toMatch(/anthropic/);
+    expect(container.textContent).toMatch(/Route/);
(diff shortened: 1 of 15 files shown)


chicoxyzzy:
@copilot all fixed. let's do a final review pass

PR #85, opened 2026-05-11
feat(release): publish updater manifest to hecate.sh, decouple auto-update from GitHub "latest"

chicoxyzzy:
## Summary

Decouple in-app auto-update from GitHub's `/releases/latest/` redirect by publishing the updater manifest to `https://hecate.sh/releases/alpha/latest.json` after each release. Sets up the long-term policy where every release stays a GitHub pre-release until `v1.0.0` — the semver suffix (`-alpha.N`) already signals pre-stability, and we want package managers, mirrors, and downstream automation that respects the flag to skip our alpha releases.

GitHub refuses to resolve `/releases/latest/` to a pre-release, which has made auto-update silently inert across the alpha cycle (it worked on alpha.27 only because we manually un-pre-released that one). This PR fixes the root cause: route auto-update through a channel we control.

## What changes

**Workflow (`.github/workflows/_tauri-shared.yml`):** new `publish-updater-website` job, runs after `publish-updater-manifest`. Downloads `latest.json` from the release, drops it under `website/public/releases/alpha/`, commits to master, dispatches `website.yml` (because pushes made with `github.token` deliberately don't trigger downstream workflows), then polls `https://hecate.sh/releases/alpha/latest.json` until the new version is live or 10 minutes elapse. Half-broken manifests fail the job loudly.

**Tauri config (`tauri/src-tauri/tauri.conf.json`):** bundled `endpoints` array now points at `https://hecate.sh/releases/alpha/latest.json`. Pubkey unchanged — signature verification keeps working for alpha.21–27 clients across the channel switch.

**Docs:**
- `docs/release.md` — new Pre-release policy section with the rationale and the historical note on alpha.27.
- `docs/desktop-updater-signing.md` — updated configuration example, expanded pipeline description, two new troubleshooting entries (manifest didn't propagate; bundle has stale endpoint).
- `docs-ai/tasks/release.md` — bridge-release procedure for alpha.28, revised CI wall-clock, clarified per-job artifact ownership.

## Transition (not in this commit)

The bridge release lands separately after this PR merges:

1. Cut alpha.28 — ships with the new endpoint baked in; CI publishes to `hecate.sh/releases/alpha/latest.json`.
2. `gh release edit v0.1.0-alpha.28 --prerelease=false --latest` so existing alpha.21–27 installs find it via their OLD baked-in endpoint and prompt the upgrade.
3. Alpha.21–27 users land on alpha.28 with the NEW endpoint baked in.
4. Alpha.29+ ship as pre-releases by default. Policy is self-enforcing from here on.

## Test plan

- [x] YAML syntax check on `_tauri-shared.yml` (parsed via `js-yaml`).
- [x] JSON syntax check on `tauri.conf.json`.
- [ ] Reviewer to sanity-check the `workflow_dispatch` trigger path against GitHub's auto-trigger-suppression rule (pushes from `github.token` don't fire downstream workflows; `workflow_dispatch` IS allowed — see [GitHub docs](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow)).
- [ ] PR CI: the `tauri-build.yml` PR-validation matrix runs the build path but not the publish jobs (those are gated on `inputs.tagName != ''`). Confirms no syntax regression in the workflow.
- [ ] Tag-driven verification: alpha.28 (post-merge bridge release) is the real test. Watch for the `publish-updater-website` job to commit `website/public/releases/alpha/latest.json` on master, dispatch `website.yml`, and pass the propagation check.

## Risk assessment

- **Failure mode if Pages deploy hangs:** the verification step caps at 10 min and fails loud. The release is otherwise complete (GitHub Release page has all bundles + manifest); only the website channel lags. Manual recovery: re-run the `publish-updater-website` job after fixing Pages, or invalidate Fastly cache.
- **Failure mode if the manifest fails to commit:** the `publish-updater-manifest` job's GitHub Release upload happens first and stands independently. Manual recovery: hand-write the manifest at `website/public/releases/alpha/latest.json` and commit.
- **Failure mode if `workflow_dispatch` is forgotten:** PRs that add new commits don't fire `website.yml`, so testing this end-to-end requires a tag push.
- **Blast radius if this misbehaves:** auto-update doesn't work for alpha.28+ users until the manifest propagates. Falls back to the user manually downloading from the Releases page (which is the current state pre-alpha.28 anyway). No data loss, no provider impact, no gateway impact.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


chicoxyzzy:
Round 2 — all five review items addressed.

**1. `actions: write` on both sides.** Reusable-workflow permissions are capped by the caller's job-level permissions, so granting `actions: write` only inside `_tauri-shared.yml` would still 403. Added at both layers:

- `.github/workflows/release.yml` — `tauri` job's `permissions:` block
- `.github/workflows/_tauri-shared.yml` — `publish-updater-website` job's `permissions:` block

The accompanying comments explicitly call out the caller/callee permission interaction so future readers don't drop the caller-side grant by accident.

**2. Unconditional dispatch.** The previous gate `if: steps.commit.outputs.changed == 'true'` was wrong for the recovery case. If a previous run committed the manifest but the Pages deploy failed or Fastly cache went stale, re-running this job would no-op the commit AND skip the dispatch — leaving the operator with no path forward short of editing master by hand. Removed the gate. The commit step still tracks `changed` (it's useful in logs), but the dispatch and verification steps always run.

**3. Non-fast-forward retry.** The commit step now wraps fetch + checkout + diff + commit + push in a 5-iteration loop with linear backoff. On a rejected push we refetch `origin/master`, replay the single-file change on top, and try again. Hard fail (CI annotation + exit 1) after 5 attempts.

**4. Path-filter wording.** Reworded in three places to make the dispatch mechanic explicit:

- `_tauri-shared.yml` — the job-level header comment now explains that `github.token` pushes don't trigger workflows on push events, and that `workflow_dispatch` is the documented workaround.
- `docs/desktop-updater-signing.md` — the pipeline description's step 2 now describes the explicit dispatch + the unconditional-run + the non-FF retry.
- `docs-ai/tasks/release.md` — both the auto-update footgun bullet and the "Watch CI" job descriptions updated.

**5. Bridge-release wording in `docs/release.md`.** The prior version conflated alpha.27 and alpha.28. Corrected story:

- alpha.27 was the last release built with the GitHub-native endpoint baked in.
- alpha.28 is the bridge. It ships with the new `hecate.sh` endpoint AND is manually flipped to non-prerelease + latest after CI.
- The flip is what makes alpha.21–27 clients' OLD baked-in endpoint resolve to alpha.28's manifest.
- Once they upgrade to alpha.28, they have the NEW endpoint baked in.
- alpha.29+ are pre-releases by default and route through `hecate.sh` only.

YAML syntax re-validated post-edit (both workflow files parse clean). Net diff on this round: +106 / −45 across 5 files.

@copilot please re-review.

PR #86, opened 2026-05-11
feat(website): release-aware download buttons on the homepage

chicoxyzzy:
## Summary

The homepage's "Download" button used to point at `https://github.com/hecatehq/hecate/releases` (the full list page) — operators had to scroll, identify their platform, and click through. Now the page reads the alpha channel's `latest.json` at build time and renders three explicit platform buttons plus a version + release-notes footer.

The plumbing is already in place from PR #85: the release workflow's `publish-updater-website` job commits `latest.json` to `website/public/releases/alpha/` on every release and dispatches `website.yml` to rebuild. So when alpha.29 ships, this page auto-updates to point at alpha.29's bundles — zero manual edits.

## Decisions worth flagging

- **Three explicit buttons over JS-based OS detection.** UA sniffing fails for ARM Linux, iPads-as-laptops, headless installs, and can't carry per-distro nuance (`.deb` vs `.AppImage`) if we ever want it.
- **macOS URL is derived, not pulled from the manifest.** The manifest carries the updater payload (`.app.tar.gz`); fresh-install operators want the `.dmg`. They share a predictable filename pattern. Comment in the code notes the Apple-Silicon-only assumption.
- **Release-notes link points at `/releases/tag/v<version>`**, not `/releases/latest`. The latter resolves to alpha.28 (the auto-update bridge release) and will keep doing so after alpha.29+ ship as pre-releases. The tag URL re-derives correctly on every build.
- **Path resolution anchored on `process.cwd()`, not `import.meta.url`.** Reviewer originally suggested `new URL(..., import.meta.url)` for robustness against cwd changes. Verified empirically that at Astro build time `import.meta.url` is the bundled chunk inside `dist/.prerender/` (e.g., `dist/.prerender/chunks/index_Cq2Fom-6.mjs`), not the source file. Relative offsets from there would couple us to Astro's internal output structure and break on Astro major-version bumps. Astro itself runs from the project root (any wrapper changing cwd would also break Astro's config + public/ resolution), so cwd is the stable anchor.
- **Try/catch with fallbacks** to `/releases`, so a fresh checkout without a published manifest still renders working links instead of failing the build.

## Test plan

- [x] `bun --bun run check` — 0 errors, 0 warnings.
- [x] `bun --bun run build` — clean.
- [x] Inspected `dist/index.html` post-build; verified the three platform URLs carry the alpha.28 version:
  ```
  Hecate_0.1.0-alpha.28_aarch64.dmg
  Hecate_0.1.0-alpha.28_amd64.AppImage
  Hecate_0.1.0-alpha.28_x64_en-US.msi
  /releases/tag/v0.1.0-alpha.28
  ```
- [ ] Reviewer to render `bun run dev` locally and sanity-check the styling of the new `.hero__release` metadata line under both desktop and mobile breakpoints. Style additions are minimal (matches existing `.eyebrow` typography family + `--muted` color).
- [ ] End-to-end test on the next release: alpha.29 should auto-update the published homepage to point at alpha.29's bundles. This is the real validation of the manifest-read + workflow-dispatch loop.

### DIFF
--- a/website/bun.lock
+++ b/website/bun.lock
@@ -9,6 +9,7 @@
       },
       "devDependencies": {
         "@astrojs/check": "^0.9.9",
+        "@types/node": "^25.6.2",
         "@typescript/native-preview": "beta",
       },
     },
@@ -245,6 +246,8 @@
     "@types/ms": ["@types/ms@2.1.0", "", {}, "sha512-GsCCIZDE/p3i96vtEqx+7dBUGXrc7zeSK3wwPHIaRThS+9OhWIXRqzs4d6k1SVU8g91DrNRWxWUGhp5KXQb2VA=="],
 
     "@types/nlcst": ["@types/nlcst@2.0.3", "", { "dependencies": { "@types/unist": "*" } }, "sha512-vSYNSDe6Ix3q+6Z7ri9lyWqgGhJTmzRjZRqyq15N0Z/1/UnVsno9G/N40NBijoYx2seFDIl0+B2mgAb9mezUCA=="],
+
+    "@types/node": ["@types/node@25.6.2", "", { "dependencies": { "undici-types": "~7.19.0" } }, "sha512-sokuT28dxf9JT5Kady1fsXOvI4HVpjZa95NKT5y9PNTIrs2AsobR4GFAA90ZG8M+nxVRLysCXsVj6eGC7Vbrlw=="],
 
     "@types/unist": ["@types/unist@3.0.3", "", {}, "sha512-ko/gIFJRv177XgZsZcBwnqJN5x/Gien8qNOn0D5bQU/zAzVf9Zt3BlcUiLqhV9y4ARk0GbT3tnUiPNgnTXzc/Q=="],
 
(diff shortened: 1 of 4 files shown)


chicoxyzzy:
Round 2 — both review concerns addressed in `<latest commit>`.

**1. Narrowed the catch to ENOENT.** Bare `catch {}` was swallowing every read/parse error. Now:

```ts
} catch (err) {
  if ((err as NodeJS.ErrnoException).code !== "ENOENT") {
    throw err;
  }
}
```

ENOENT (fresh checkout, no release published yet) stays silent — that's the legitimate bootstrap case. Corrupt JSON, schema mismatches, permission issues, anything else re-throws and fails the build with the original error message. Verified by writing invalid JSON to the manifest:

```
[ERROR] SyntaxError: JSON Parse error: Unexpected identifier "not"
error: script "build" exited with code 1
```

**2. Hid the version + release-notes line on the fallback path.** New `manifestLoaded` flag, set to `true` only inside the try block after the manifest parses successfully. The `<p class="hero__release">` paragraph now renders conditionally:

- **Manifest loaded:** `v0.1.0-alpha.28 · release notes · all releases · GitHub`
- **Bootstrap (ENOENT):** `all releases · GitHub` — no fabricated version, no misleading "release notes" link to the index.

Verified by moving the manifest aside and rebuilding — all four download buttons fall back to `/releases`, the version tag and release-notes link disappear from the footer.

**Three states tested:**

| Case | Behavior |
|---|---|
| Manifest present | All four URLs alpha.28-specific, full footer |
| Manifest absent (ENOENT) | All buttons → `/releases`, version line hidden |
| Manifest corrupt | Build fails loud with the JSON parse error |

@copilot please re-review.

PR #87, opened 2026-05-11
fix(adapters): rewrite Claude Code auth error + one-click setup deep-link

chicoxyzzy:
## What an operator sees today

When Claude Code can't sign in, the chat shows:

> Claude Code needs adapter-visible authentication: set `CLAUDE_CODE_OAUTH_TOKEN` from `claude setup-token`, `ANTHROPIC_API_KEY`, or `ANTHROPIC_AUTH_TOKEN` for the adapter environment. Restart Hecate after changing environment credentials.

Five things wrong with that for a non-developer:

1. **Jargon** — "adapter-visible authentication", "adapter environment" mean nothing without context.
2. **Three options at the same priority** — paralyzing, no recommendation.
3. **"Restart Hecate"** as the headline, but the easiest fix doesn't require a restart.
4. **No mention** that Hecate already has a guided setup card in `Settings → External agents` that takes the token via paste (Hecate stores it, injects it into the adapter, no restart).
5. **Conflates two different keys** — operators who configured Anthropic in the Providers tab reasonably think Claude Code "just works." It doesn't, because Claude Code is a separate subprocess with its own credentials. The error message gives them no hint that this is a different surface.

## What this PR changes

**Server-side (Tier 1) — friendly message + stable marker:**

```
Claude Code isn't signed in. This is a separate credential from
the Anthropic key in the Providers tab — Claude Code runs as its
own program. Click the button below to paste a token from
`claude setup-token` into the guided setup card. No restart
needed. (claude_code_auth_required)
```

The trailing `(claude_code_auth_required)` is a load-bearing token the chat UI pattern-matches on. It's stripped from the visible copy before rendering, so the operator sees clean prose. Documented at the call site in `internal/agentadapters/auth_status.go`.

Dashboard hints (`auth_status.go`) and probe-side hints (`probe.go`) get the same treatment: lead with the guided setup card, drop the three-env-var enumeration. The card itself documents the env-var alternative for power users.

**Client-side (Tier 2) — inline recovery button:**

The chat's failed-run notice now renders an **"Open Claude Code setup"** button when the failure carries the `claude_code_auth_required` marker and the message's adapter is `claude_code`. Click:

- Sets `localStorage.hecate.settingsTab = "external_agents"` so SettingsView mounts on the right sub-tab.
- Sets `sessionStorage.hecate.settingsFocus = "claude-code-guided-setup"` so the External Agents tab scrolls-and-flashes the guided setup card on mount.
- Calls `onNavigate("settings")` to switch workspaces.

One click from "chat failed" to "setup card focused, ready to paste."

The flash animation lives in `styles.css` (2.2s teal ring pulse) next to the design tokens.

## Verification

- `bun run typecheck` — clean.
- `bun run test` — 707/707 (was 704; +3 new tests on `TranscriptMessageRow`: marker stripping, button rendering + onClick, negation for cancelled runs).
- `go test ./internal/agentadapters/... -count=1` — 213/213.

## Risk

- **Marker drift.** If the server-side string is reworded and the `(claude_code_auth_required)` token is accidentally dropped, the UI button silently disappears. Mitigated by an explicit test in `registry_test.go` that asserts the token is present, plus comments at both ends of the contract.
- **Storage unavailable.** `localStorage` / `sessionStorage` access wrapped in try/catch — navigation still works without auto-scroll if storage is denied (private-browsing context, Tauri webview restrictions).
- **Other adapters with similar issues.** The marker convention `(<adapter>_auth_required)` is forward-compatible — Codex / Cursor could use the same pattern. Not implemented in this PR; Claude Code was the reported one.

### DIFF
--- a/internal/agentadapters/auth_status.go
+++ b/internal/agentadapters/auth_status.go
@@ -37,11 +37,15 @@ func DetectAuthStatus(adapter Adapter) (string, string) {
 // the chat when a Claude Code agent run fails because the adapter
 // couldn't sign in. Two priorities behind the wording:
 //
-//  1. The chat UI pattern-matches `claude_code_auth_required` (the
-//     trailing token in the parenthetical below) to render an inline
+//  1. ui/src/features/chats/ChatView.tsx pattern-matches the
+//     `claude_code_auth_required` token in this string (via
+//     .includes()) to decide whether to render an inline
 //     "Open Claude Code setup" button that deep-links to the guided
-//     setup card in Settings → External agents. Keep that token in
+//     setup card in Settings → External agents. Keep the token in
 //     the string verbatim — the UI handler depends on it.
+//     (TranscriptMessageRow only strips the trailing parenthetical
+//     marker from the visible copy; the match-and-render-button
+//     decision lives in ChatView.)
 //  2. Distinguish this credential from the Anthropic key in the
 //     Providers tab. Operators who paste a key into Providers tab
 //     reasonably expect Claude Code to "just work" — it doesn't,
--- a/internal/agentadapters/auth_status.go
+++ b/internal/agentadapters/auth_status.go
@@ -20,9 +20,9 @@ func DetectAuthStatus(adapter Adapter) (string, string) {
 			return AuthStatusOK, ""
 		}
 		if fileAny("${HOME}/.claude.json", "${HOME}/.claude/settings.json", "${HOME}/.claude/.credentials.json") {
-			return AuthStatusUnknown, "Claude Code config is present, but the ACP adapter may still need adapter-visible auth. Use Test adapter; if it fails, " + claudeCodeACPAuthHint()
+			return AuthStatusUnknown, "Claude Code config is present on disk, but the ACP adapter may need its own token. Use Test adapter to check; if it fails, open the guided setup card below."
 		}
-		return AuthStatusUnknown, "Use Test adapter; if Claude Code reports auth errors, " + claudeCodeACPAuthHint()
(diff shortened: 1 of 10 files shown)


chicoxyzzy:
Round 2 — all three Copilot concerns addressed in `<latest commit>`:

**1. Focus-target injection (`SettingsView.tsx`).** Added a `KNOWN_FOCUS_TARGETS` allowlist (`Set<string>` containing only `"claude-code-guided-setup"` for now). The sessionStorage value is validated against the set before it touches a DOM query, so a malformed or unexpected token just gets ignored. Producer and consumer are both internal, so this is the right shape — extending to a new target is "add to the set + add to the producer," no escaping logic needed. Comment in the code explains how to extend.

**2. Leaky inner timer (`SettingsView.tsx`).** The 2.2s class-removal timeout was being created inside the outer 0ms deferred timeout but never tracked. Both handles are now stored and cleared in the effect's cleanup, so an unmount mid-flash neither leaks the timer nor runs `classList.remove` on a detached node.

**3. Misleading "UI handler" pointers (`auth_status.go` + `registry_test.go`).** Both comments said `TranscriptMessageRow` is where the `claude_code_auth_required` marker is matched. The actual `.includes()` call lives in `ChatView.tsx`, which builds the `setupAction` prop and passes it down. `TranscriptMessageRow` only strips the marker from the visible message text. Both comments now name `ChatView.tsx` explicitly with the file path, and call out the responsibility split so future readers don't wander to the wrong file.

**Verification:** `bun run typecheck` clean, `bun run test` 707/707, `go test ./internal/agentadapters/... -count=1` 213/213. No new tests this round — the existing `TranscriptMessageRow.test.tsx` coverage (marker stripping + button render + cancelled-state negation) still pins the behavior; the allowlist + timer-cleanup are internal robustness changes that don't add new observable surfaces to test.

@copilot please re-review.

PR #89, opened 2026-05-11
fix(tauri): stamp packaged sidecar versions

chicoxyzzy:
## Summary

Packaged Tauri builds were stamping the native app version, but the embedded Go sidecars were still built with plain `go build`. That left `internal/version.Version` at its default `dev`, so the gateway `/healthz` response and UI status bar showed `dev` inside released desktop bundles.

This PR adds a dedicated `just tauri-build-sidecars` recipe that resolves the same release version used by Tauri and injects it into both `hecate` and `hecate-acp` with Go ldflags. The Tauri release matrix now uses that recipe before staging the platform-specific `externalBin` files.

## Verification

- `TAURI_VERSION=v1.2.3-alpha.4 just tauri-build-sidecars`
- `./hecate --version` → `1.2.3-alpha.4`
- `./hecate-acp --version` → `1.2.3-alpha.4`


chicoxyzzy:
@copilot 

What changed:

- Added `scripts/resolve-tauri-version.ts` as the shared version resolver.
- `scripts/stamp-version.ts` now uses the shared resolver instead of duplicating env/tag/Cargo parsing.
- `Justfile` now uses the same resolver for `tauri-build-sidecars`.
- `_tauri-shared.yml` sets `TAURI_VERSION` at the job level, so both sidecar build/staging and Tauri metadata stamping receive the same value.
- `tauri-build.yml` now triggers when the shared resolver changes.
- Tauri docs now describe `just tauri-sidecar <target>` as the build+stage command, update the `GOEXE` staging behavior, and clarify that `TAURI_VERSION` may be v-prefixed.

PR #91, opened 2026-05-12
feat(agent-adapters): improve Claude Code onboarding

chicoxyzzy:
## Summary

This PR makes Claude Code onboarding much harder to misread by separating three states that were previously blurred together:

- whether the adapter command can be discovered/launched
- whether Claude Code has adapter-visible auth that Hecate can inject
- whether the operator has saved a valid Claude Code OAuth token in Hecate

The visible result is a guided Claude Code setup flow in Chats and Settings instead of a misleading ready/auth label. Settings keeps the token editor available for update/removal, validates token shape before saving, and reports saved-token state separately from adapter readiness. Chats now keep the External Agent setup screen visible until Claude Code has credentials Hecate can actually pass to the ACP adapter.

## Details

- Adds Claude Code setup helpers for CLI/npx detection and setup-command status.
- Extends adapter health/readiness responses with Claude setup and saved-token state.
- Validates `CLAUDE_CODE_OAUTH_TOKEN`-style values before saving them into the local secret store.
- Keeps Anthropic provider API keys separate from Claude Code adapter tokens in Settings copy and UI state.
- Refreshes External Agent onboarding copy for Claude Code, Codex, and Cursor so each adapter gets the right install/auth guidance.
- Adds `GATEWAY_AGENT_ADAPTER_DISCOVERY_OVERRIDES` plus `just dev-no-agent-adapters` / `just dev-agent-adapters` for manual smoke tests and e2e fixtures. This override is discovery-only: it changes Settings/Chats readiness surfaces but does not create fake runnable adapter processes.
- Documents the override in development, runtime API, and external-agent docs so manual smoke testing and e2e usage are discoverable.

## Tests

- `go test ./internal/agentadapters ./internal/api`
- `cd ui && bun run test -- ChatView.test.tsx SettingsView.test.tsx api.test.ts`

## Notes for reviewers

The discovery override intentionally does not make a forced-available adapter runnable. It is only for reproducing UI states like “no adapters installed” without uninstalling local tools. The API test `TestAgentAdaptersHonorsDiscoveryOverride` locks that `/hecate/v1/agent-adapters` reflects the override without relying on local machine state.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -220,6 +220,12 @@ GATEWAY_AGENT_CHAT_IDLE_TIMEOUT=0s
 # Claude Code when direct adapter binaries are not on PATH. Empty uses the
 # platform user-cache location.
 HECATE_AGENT_ADAPTERS_DIR=
+# Development / e2e-only discovery override for external-agent adapter
+# readiness. Comma-separated adapter states: "all=missing",
+# "claude_code=missing,codex=available", etc. Supported states are
+# "missing" and "available". This only affects discovery surfaces in
+# Settings/Chats; it does not create fake runnable adapter processes.
+GATEWAY_AGENT_ADAPTER_DISCOVERY_OVERRIDES=
 
 # Rate limiting — process-local token bucket on /v1/chat/completions
 # and /v1/messages. Off by default. When enabled, every response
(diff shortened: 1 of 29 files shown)


chicoxyzzy:
@copilot-pull-request-reviewer please review again. I pushed fixes for the open review threads and resolved the addressed conversations.

PR #92, opened 2026-05-12
feat(chats): unify agent controls and setup

chicoxyzzy:
## Summary

- Rework Chats around a single sidebar agent picker for Hecate, Codex, Claude Code, and Cursor, with Hecate first and external agents fixed per chat once created.
- Add ACP session preparation/config-option plumbing so adapter controls can be shown before the first prompt and real ACP config options can be applied once the session exists.
- Improve external-agent setup and diagnostics: Claude Code guided setup, adapter readiness/auth affordances, local-provider onboarding copy, workspace dialog cancel handling, richer activity details, raw-output grouping, and inspect/revert actions for captured diffs.
- Split large ChatView setup surfaces into focused components for agent controls, Claude Code setup, and chat instructions. Add shared brand avatar/dropdown primitives and wire Hecate Chat instructions into both direct-model and task-backed turns.
- Refresh docs and screenshots for the updated chat/settings/provider surfaces.

## Verification

- cd ui && bun run typecheck
- cd ui && bun run test
- go vet ./internal/api ./internal/agentadapters ./internal/agentchat
- go test ./internal/api ./internal/agentadapters ./internal/agentchat

## Notes for reviewers

- The PR intentionally touches both backend ACP session state and React chat architecture because the UI now needs early adapter/session controls before the first message.
- External agent selection is now a chat-creation choice rather than a mutable active-session toggle, which keeps transcript history and adapter-specific controls predictable.


chicoxyzzy:
@copilot addressed your comments. round 2?

chicoxyzzy:
@copilot addressed the latest comments:

- Agent Chat memory-store snapshots now deep-copy `ConfigOptions`, including nested select options, so callers cannot mutate stored session state through Create/Get/List results.
- ACP tool-summary truncation is now rune-safe for UTF-8 text.
- `AgentChatConfigOptionRecord.type` keeps literal narrowing for `select` / `boolean` while still allowing future string values.

Verified with:
- `go vet ./internal/agentchat ./internal/agentadapters ./internal/api`
- `go test ./internal/api ./internal/agentchat ./internal/agentadapters`

Please take another look.

chicoxyzzy:
Follow-up: the previous push exposed a CI-only `go test -race ./...` failure in `TestDetectVersionParsesStderr`. I replaced the brittle shell stderr fixture with a tiny compiled Go helper binary, so the test still validates stderr parsing but no longer flakes under race-mode subprocess load.

Additional verification:
- `go vet ./internal/agentadapters`
- `go test ./internal/agentadapters`
- `go test -race -timeout 10m ./internal/agentadapters`

@copilot please review the latest head when you get a chance.

chicoxyzzy (commented):
Spent some time reading through the new package + the create-session flow alongside Copilot's notes. Most of what's worth flagging Copilot already caught (the `cloneSession` shallow-copy of `ConfigOptions`, the `acpSession.setConfigOptions`/`configOptionsSnapshot` shallow snapshots, the `trimToolSummary` UTF-8 byte-split, the TS union widening on `AgentChatConfigOptionRecord.type`, the `PrepareSession`-then-`UpdateSession` state desync, the `formatBytes(0)` falsy guard, the `BrandAvatar` aria/size issues — agree with all of those).

Two issues I'd add on top, and two Copilot false positives worth flagging so the author doesn't burn time chasing them:

## New comments (left inline)

1. **`internal/agentcontrols/config_options.go` — `FromACPOptions` silently drops unknown option variants.** When ACP adds a new variant beyond `Select` / `Boolean`, the switch falls through and the option vanishes from the output without any signal. Worth either logging the unknown variant or panicking in tests so we notice when the SDK adds something.

2. **`internal/agentcontrols/config_options.go` — `BuildACPSetRequest` silently prefers `BoolValue` over `Value`.** If a caller supplies both (e.g. a request handler that doesn't fully normalize input), the boolean path wins and the string value is dropped without an error. Should either fail loud on over-supply or document the precedence in the doc comment.

## Copilot false positives (no action needed)

- **"`PrepareSession` is wrapped with `agentChatTimeout` (30 minutes)"** — actually `agentChatPrepareTimeout = 10 * time.Second` (handler_agent_chat.go:24, used at line 145). The 30-minute `agentChatTimeout` is for actual runs, not session prep. Copilot misread the constant name.

- **"In `HandleCreateAgentChatSession`, when `h.agentChatRunner` is nil you create a new SessionManager just for `PrepareSession`, but that runner is not stored anywhere"** — actually handler_agent_chat.go:107-110 returns HTTP 500 immediately when the runner is nil, with no fallback SessionManager creation. Copilot seems to have confused this flow with a different handler's fallback.

(The other Copilot note about the same handler — "`HandleSetAgentChatConfigOption` calls `SetSessionConfigOption` without guarding for a nil runner" — is real and worth fixing.)

## Things that look good

- `agentcontrols` is a clean projection layer — keeps the persisted shape independent of the SDK's generated union internals (called out in the doc comment, agree with the reasoning).
- The new e2e test (`chat.spec.ts:100` "New chat creates an external-agent session with controls before the first prompt") covers the create-session-with-config-options happy path including the `config_options` payload landing in the response — good regression pin.
- The cleanup-on-prepare-failure path in `HandleCreateAgentChatSession` (handler_agent_chat.go:154 `h.agentChat.Delete`) is the right shape; pair it with `h.agentChatRunner.CloseSession` in the `UpdateSession`-failure branch (per Copilot's note) and the state machine closes cleanly.

Not blocking — happy to approve once the load-bearing Copilot notes (nil-runner guard on `SetSessionConfigOption`, deep-copy of `ConfigOptions`, `PrepareSession`+`UpdateSession` state cleanup) plus the two above land.

internal/agentcontrols/config_options.go
```diff
@@ -0,0 +1,155 @@
...
+		return nil
+	}
+	out := make([]ConfigOption, 0, len(options))
+	for _, option := range options {
+		switch {
+		case option.Select != nil:
+			out = append(out, fromACPSelect(*option.Select))
+		case option.Boolean != nil:
+			out = append(out, fromACPBoolean(*option.Boolean))
+		}
```
chicoxyzzy, line 61: `FromACPOptions` silently drops options where neither `Select` nor `Boolean` is set — the switch falls through with no entry written to `out` and no log. When ACP adds a new variant (the SDK's union shape suggests more variants are possible), the option vanishes from the UI without operator-visible signal. Suggest one of: (a) log the unknown variant via a package-level logger, (b) include an `unknown` entry in the output with the raw type name so the UI can render a placeholder, or (c) at minimum a test that asserts an error/panic on the unknown shape so we notice in CI when the SDK gains a new variant.

internal/agentcontrols/config_options.go
```diff
@@ -0,0 +1,155 @@
...
+	if req.Value == "" {
+		return acp.SetSessionConfigOptionRequest{}, fmt.Errorf("value is required")
+	}
+	return acp.SetSessionConfigOptionRequest{
+		ValueId: &acp.SetSessionConfigOptionValueId{
+			SessionId: acp.SessionId(req.SessionID),
+			ConfigId:  acp.SessionConfigId(req.ConfigID),
+			Value:     acp.SessionConfigValueId(req.Value),
+		},
+	}, nil
```
chicoxyzzy, line 98: Silent precedence: when a caller supplies both `BoolValue` (non-nil) AND `Value` (non-empty), the boolean branch wins and `Value` is discarded without any error. Validation at line 77 only fires when *both* are empty. Two ways out: (a) reject over-supplied requests explicitly (`if req.BoolValue != nil && req.Value != "" { return ..., fmt.Errorf("provide either Value or BoolValue, not both") }`), or (b) document the precedence in the doc comment so callers know what to expect. Either is fine — silent dropping is the worst of the three options.

chicoxyzzy:
@copilot addressed the latest review comments:

- ACP session config options now deep-copy nested select options both when storing adapter-provided options and when returning snapshots.
- External-agent session creation now cleans up if persistence fails after `PrepareSession`: it closes the prepared runner session and deletes the partially-created chat record.
- Zero-byte output activity rows now render `0b` instead of hiding the size.
- Unboxed unknown-brand avatars now keep fixed dimensions so custom provider/agent icons stay aligned.

Verification:
- `go vet ./internal/agentadapters ./internal/api`
- `go test ./internal/agentadapters ./internal/api`
- `go test -race -timeout 10m ./internal/agentadapters`
- `cd ui && bun run test -- TranscriptActivityTimeline.test.tsx ui.test.tsx`

Please take another look.

internal/agentcontrols/config_options.go
chicoxyzzy: Updated in `f73766bb`: unknown ACP config option variants are now preserved as explicit `unknown` placeholders instead of being skipped. I also added tests for unknown-only and unknown-alongside-known options so future ACP union additions stay visible in CI/UI.

internal/agentcontrols/config_options.go
chicoxyzzy: Updated in `f73766bb`: `BuildACPSetRequest` now rejects requests that provide both `Value` and `BoolValue` before constructing the ACP wire request, with validation coverage for the over-supplied case.

chicoxyzzy:
Addressed the latest review comments in `f73766bb` and resolved the threads.

What changed:
- Unknown ACP config option variants now become explicit `unknown` placeholders instead of disappearing from the normalized option list.
- Config option updates now reject over-supplied requests that include both `Value` and `BoolValue`.
- Added tests for unknown ACP variants and the mutual-exclusion validation path.

Verified with:
- `go vet ./internal/agentcontrols ./internal/agentadapters ./internal/api`
- `go test ./internal/agentcontrols ./internal/agentadapters ./internal/api`
- `go test -race -timeout 10m ./internal/agentadapters`
- `cd ui && bun run test -- TranscriptActivityTimeline.test.tsx ui.test.tsx`

Please take another look when you get a chance.

PR #97, opened 2026-05-13
fix(readiness): unify provider repair states

chicoxyzzy:
## Summary

- Unify chat setup and provider readiness repair states so blocked Hecate Chat sends point operators to Connections with the same root cause shown in provider setup.
- Add actionable provider repair hints for adding providers, opening blocked provider details, and refreshing provider/model discovery.
- Surface the same readiness guidance in Connections, Providers, and chat setup states, with backend/API tests for important repair reasons.

## Details

This keeps the provider readiness contract as the canonical source for setup failures and teaches the UI how to turn that contract into safe next actions. Operators now see more specific guidance for missing credentials, no discovered models, disabled providers, self-referential local routes, and backend-reported routing blockers.

The PR also expands tests around those contracts and repair paths so future provider-readiness changes do not drift between backend responses, Connections copy, Providers actions, and Chat setup CTAs.

## Verification

- `just verify`

### DIFF
--- a/docs-ai/skills/ui/SKILL.md
+++ b/docs-ai/skills/ui/SKILL.md
@@ -129,6 +129,10 @@ Each section has exactly one job: orient, inspect, compare, edit, or confirm. If
   to a dead-end warning. If the backend provides a suggested replacement model,
   expose it as an action and reset the provider route to "All providers" before
   selecting it, because the suggested model may belong to a different provider.
+- Chat send blockers should flow through `resolveChatSetupRepairState` in
+  `ui/src/lib/chat-setup-readiness.ts`. Keep the empty state, composer notice,
+  and disabled-send copy aligned there instead of adding one-off branches to
+  `ChatView`.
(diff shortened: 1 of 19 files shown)


chicoxyzzy:
Addressed the review feedback in ce06ec5a.

What changed:
- Suppressed duplicate empty-chat repair notices while keeping the tools-disabled composer action where the capability-write busy state lives.
- Let Claude Code setup show as a composer repair for non-empty sessions so blocked sends always have visible guidance.
- Centralized provider repair action labels in `provider-readiness.ts`.
- Tightened the null-repair readiness summary edge case.
- Stopped using runtime provider names as configured provider IDs; runtime-only cases now fall back to refresh guidance.

Verified with:
- `cd ui && bun run test -- provider-readiness.test.ts ProvidersView.test.tsx ChatView.test.tsx`
- `cd ui && bun run test`

@copilot-pull-request-reviewer please review again.

PR #101, opened 2026-05-14
feat(llamacpp): Hecate-managed local models — download + run via llama.cpp

chicoxyzzy:
## Summary

Adds first-party support for downloading and running local models on the bundled `llama-server` binary, similar to Ollama / LM Studio. The feature is engine-agnostic at the URL level (`/hecate/v1/local-models/*`) so future MLX or other runtimes share the namespace.

Lands behind a feature gate — when `HECATE_LLAMA_SERVER_BIN` is unset (or `HECATE_LOCAL_MODELS` not opted in), the surface stays dormant: handlers return a useful 503 with `local_models_unavailable`, no provider is auto-registered, no child processes spawn.

### v1 — backend foundation, UI surface, Tauri sidecar bundling
- **Backend** (`internal/llamacpp/`): typed Installer (sha-verified, atomic rename, cancellable), Runtime (single-child supervisor with crash listener), Proxy (gateway-internal reverse proxy with `FlushInterval=-1` for streaming), Service facade composing them plus a curated Catalog.
- **HTTP API**: `GET /catalog`, `GET /installed`, `POST /install` + SSE `events`, `DELETE` install/installed, `GET /runtime`, `POST runtime/start|stop`, plus the internal `/hecate/internal/llamacpp/v1/{path...}` proxy mount.
- **`/v1/models` integration**: installed local models join the chat composer's model dropdown alongside provider-routed models.
- **Auto-registered provider**: on boot, a managed `llamacpp` provider row is created pointing at the internal proxy. Operator overrides (matched by `PresetID="llamacpp"`) are left alone — surfaces as `ErrAutoProviderOperatorOwned` for log-warning visibility.
- **UI** (`ui/src/features/providers/`): LocalModelsCard summary in the Connections panel, LocalModelsSlideOver with Runtime / Installed / Catalog / Custom URL sections. SSE-driven install progress, single confirm-modal-gated Uninstall, per-row Start with hidden Start on the loaded row.
- **Tauri**: `llama-server` joins `claude-code` and `codex` as a sidecar; resolved binary path is environment-wired into the gateway at startup.

### v2 — lazy-download, LRU keep-warm, gated repos, HuggingFace browse
1. **Headless lazy-download** — `HECATE_LOCAL_MODELS_LAZY_DOWNLOAD=on` triggers a one-time download of `llama-server` from the pinned upstream llama.cpp release at boot. Atomic rename + chmod, sha-verified when the pin carries a digest, cached at `<data_dir>/llamacpp/bin/`. Tauri builds leave it OFF — the bundled binary is the source of truth there.
2. **LRU keep-warm** — `HECATE_LOCAL_MODELS_MAX_RESIDENT=N` keeps the N most-recently-used models loaded across requests; the (N+1)th `EnsureLoaded` evicts the coldest session. Default 1 preserves v1 single-child behavior. All sessions race-clean.
3. **Gated HuggingFace repos** — per-install token via `InstallSpec.hf_token`, env fallback via `HUGGINGFACE_TOKEN`. The installer attaches `Authorization: Bearer <token>` only when present. Token is not persisted at rest; v2 deliberately scopes to single-use.
4. **HuggingFace browse + search** — server-side proxy at `GET /local-models/huggingface/search` and `.../repos/{owner}/{name}`. Search pins `filter=gguf`, sort by downloads. File tree returns only `.gguf` files with LFS sha256 + size + canonical resolve URL. Browser never sees the HF token. UI surfaces a "Browse HuggingFace" section between Catalog and Custom URL with search → result rows (gated badge, download count) → per-repo file expander → Install drops into the existing install flow with sha256 attached.

### Telemetry, observability, error mapping
- New per-feature spans: `local_model.install`, `local_model.runtime`, `local_model.proxy` join the central span table with `hecate.phase` labels — `TestAllTelemetryEventsHaveSpecificSpanAndPhase` covers them.
- Stable error codes: `local_models_unavailable`, `local_model_not_installed`, `local_model_runtime_unavailable`, `local_model_install_already_running`, `local_model_install_not_found`, `huggingface_gated`, `huggingface_not_found`, `huggingface_upstream_error`.
- OTel install events: `local_model.install.{started,progress,completed,failed,cancelled}` with byte counts + sha256 attributes. Runtime crash events surface with reason + uptime.

### Docs
- `docs/rfcs/local-models-llamacpp.md` — full RFC: scope, architecture, error mapping, boot reconciliation contract, operator override path, v1/v2 boundary.
- `docs/local-models.md` — operator guide: availability matrix, env knobs, HTTP API, error codes, troubleshooting recipes.
- `.env.example` — `HECATE_LLAMA_SERVER_BIN`, `HECATE_LOCAL_MODELS`, `HECATE_LOCAL_MODELS_LAZY_DOWNLOAD`, `HECATE_LOCAL_MODELS_MAX_RESIDENT`, `HUGGINGFACE_TOKEN`.

## Stats

52 files changed, +11,940 / −4. ~1880 backend tests green under `-race`; 734 UI tests green.

## Test plan

- [ ] Backend: `go test -race ./internal/llamacpp/... ./internal/api/... ./internal/profiler/... ./internal/telemetry/...`
- [ ] Full race suite: `go test -race ./...`
- [ ] UI: `bun run typecheck && bun run test` (vitest, 35 files / 734 tests)
- [ ] Smoke: dormant gateway returns `local_models_unavailable` 503 on `/catalog`
- [ ] Smoke: with `HECATE_LLAMA_SERVER_BIN` set, `/catalog` returns curated list with `installed=false`
- [ ] Smoke: install a catalog model → SSE progresses → registry row appears → `/v1/models` lists it
- [ ] Smoke: Start → llama-server child boots → proxy routes a chat request → Stop cleans up
- [ ] Smoke: HF search returns gguf-filtered repos; file expander lists `.gguf` files with sha; Install path uses LFS sha256
- [ ] Tauri build: `pnpm tauri build` produces a bundle that ships `llama-server` as a sidecar
- [ ] Headless lazy-download path: gateway with no `LLAMA_SERVER_BIN` but `LAZY_DOWNLOAD=on` downloads the pinned binary to `<data_dir>/llamacpp/bin/` on first start


chicoxyzzy:
All 13 Copilot review comments addressed across [`e91d4622`](https://github.com/hecatehq/hecate/commit/e91d4622), [`f7491294`](https://github.com/hecatehq/hecate/commit/f7491294), and follow-ups. Resolution per comment:

| # | Location | Fix |
|---|---|---|
| 1 | `internal/llamacpp/runtime.go:625` Stop hang on pre-watcher | `watcherDone` stays nil until the watcher goroutine is actually spawned; Stop's receive skips nil channels. Regression test `TestRuntime_StopWhilePreWatcher` reproduces the race (`e91d4622`). |
| 2 | `internal/llamacpp/proxy.go:82` GET /v1/models 400s | Added `methodHasJSONBody` + `isBodylessPath` predicates and a `forwardBodyless` path that routes through `ProxyRuntime.ActiveBaseURL`. Tests cover happy path, "nothing loaded" 503, and large body (`e91d4622`). |
| 3 | `ui/src/lib/api.ts:333` HF token in URL | Token now rides `Authorization: Bearer` end-to-end; handler reads from header with HUGGINGFACE_TOKEN env fallback; query-string token is explicitly ignored (regression test in `handler_local_models_test.go`). |
| 4 | `scripts/fetch-llama-server.ts:70` unpinned sha | Script fails closed when `sha256` is empty; `HECATE_ALLOW_UNVERIFIED_LLAMA_SERVER=1` is the documented dev-only bypass (prints observed digest for recording). Backfilled `b4404` digest in `f7491294`. |
| 5 | `internal/llamacpp/binary_resolver.go:125` resolver no sha | `Resolve()` returns `ErrBinarySHARequired` when `AssetSHA256` is empty unless `AllowUnverifiedDownload` (test-only). `DefaultBinarySpec` now carries the pinned darwin/arm64 digest. |
| 6 | `tauri/src-tauri/src/sidecar.rs:164` debug resolution | Debug branch now checks `tauri/src-tauri/binaries/llama-server-<triple>` first (where the fetch script stages), with the historical repo-root fallback after. |
| 7 | `docs/local-models.md:341` stale out-of-scope | Section replaced with a Status block reflecting v1+v2 implemented; "Still out of scope" trimmed to Linux/Windows bundles, per-model fine config, auto-update, non-llama.cpp engines. |
| 8 | `docs/rfcs/README.md:28` "Not implemented" | Updated to mark v1+v2 implemented with the four v2 features called out. |
| 9 | `cmd/hecate/main.go:338` non-runtime handlers when binary missing | `localModelsService()` gate now also checks `FeatureAvailability().Available`; catalog/installed/install/HF endpoints return 503 when the binary is unresolved. `/runtime` keeps emitting the dormant introspection body via `localModelsServiceForIntrospection`. |
| 10 | `internal/llamacpp/proxy.go:143` 64 KiB body cap | Raised to 16 MiB matching the gateway's accepted chat payload size. Test `TestProxy_AcceptsLargeRequestBody` exercises a 1 MiB payload. |
| 11 | `internal/llamacpp/service.go:271` Provider.ID not set | Managed row now has explicit `ID="llamacpp"` (`managedProviderID` constant) so `owned_by` in `/v1/models` matches the registered provider. |
| 12 | `internal/llamacpp/service.go:236` stale URL on reboot | `EnsureAutoRegisteredProvider` distinguishes managed (ID match) from operator-override (different ID, same PresetID). Managed row gets re-upserted on each boot to refresh `BaseURL`. Regression test `TestService_EnsureAutoRegisteredProvider_RefreshesManagedRow`. |
| 13 | `internal/llamacpp/installer.go:510` stale gated msg | 401/403 now points at the v2 recovery paths: missing token → set `hf_token` or `HUGGINGFACE_TOKEN`; rejected token → verify access + expiry. |

Backend race suite (1889 tests) and UI suite (734 tests) both green. The Tauri matrix is also running for the first time on this PR — `_tauri-shared.yml`'s `publish-updater-website` job declared `contents: write, actions: write`, which `tauri-build.yml` capped at `contents: read`, causing startup_failure for every run since `35b59cd4` (2026-05-01). Lifted the cap in `e7a3643f` with the security tradeoff documented inline.

chicoxyzzy:
All three findings addressed in [`bf048da4`](https://github.com/hecatehq/hecate/pull/101/commits/bf048da4):

| # | Finding | Fix |
|---|---|---|
| P1 | `Stop()` lost the concurrently-starting child — promoted session no longer in pool, child leaked | [internal/llamacpp/runtime.go](internal/llamacpp/runtime.go) checks `r.sessions[modelID] == session` at both lock reacquisitions in `EnsureLoaded` (after `starter.Start` and after health polling). If Stop dropped the session under us, kill the orphaned handle and return `ErrRuntimeNotRunning`. Releases the mutex during `handle.Stop` so `stopTimeout` doesn't pin the runtime; re-locks for the outer `defer r.mu.Unlock()`. Regression test `TestRuntime_StopDuringSpawnKillsOrphanedChild` uses a `blockingStarter` to keep the race window open and asserts `handle.stopCount == 1` plus empty session pool after the race. |
| P1 | Headless auto-register stored relative `BaseURL` when `PublicURL` empty | [cmd/hecate/main.go](cmd/hecate/main.go) defers `EnsureAutoRegisteredProvider` until after `net.Listen` returns, so it can fall back to `"http://" + listener.Addr().String()` when `cfg.Server.PublicURL` is empty. Tauri still passes `PublicURL` explicitly; headless gateways enabled via `HECATE_LLAMA_SERVER_BIN` now get an absolute URL too. |
| P2 | Placeholder `llama-server` stubs looked real to both sides | Justfile + CI workflow write `# hecate-llama-server-placeholder` sentinel into the stub. [tauri/src-tauri/src/sidecar.rs](tauri/src-tauri/src/sidecar.rs)'s new `is_llama_server_placeholder` scans the first 512 bytes and rejects any match — `HECATE_LLAMA_SERVER_BIN` is never set to a stub. [internal/llamacpp/service.go](internal/llamacpp/service.go) does the same check in `FeatureAvailability` for the headless path where an operator might point at the stub directly — returns `Reason: "binary_is_placeholder"`. Regression test in [service_test.go](internal/llamacpp/service_test.go). |

**Verified locally**: 1891 Go tests pass under `-race`, 734 UI tests pass, 7 Rust tests pass, `actionlint` clean on all three workflows.

Thanks for the careful review — the orphan-child case in particular was the kind of bug that would have been very confusing to debug live (runtime says idle, but `ps` shows a llama-server running).

PR #102, opened 2026-05-15
feat(ui): unify chat shell and agent setup

chicoxyzzy:
## Summary

Unifies the chat workspace around a single agent-aware shell for Hecate Chat and external agents.

- Creates Hecate and external-agent chats consistently before the first prompt, with agent selection, shared setup/repair states, and provider/model readiness copy.
- Adds agent-chat rename support via `PATCH /hecate/v1/agent-chat/sessions/{id}` and wires sidebar rename through the UI.
- Polishes chat setup surfaces: Claude Code guided setup, local provider onboarding, dropdown behavior, full workspace paths, chat notices, changed-file actions, and shared transcript activity details.
- Prevents Observability from flashing the empty traces state while the initial traces request is still loading.
- Updates docs, screenshot fixtures, and RFC notes for the chat runtime UX consistency work.

## Validation

- `go test ./...`
- `go vet ./...`
- `go test -tags e2e ./e2e/...`
- `cd ui && bun run test`
- `cd ui && bun run test:e2e`


chicoxyzzy (commented):
Large, well-scoped UI redesign. The RFC at `docs/rfcs/chat-runtime-ux-consistency.md` cleanly captures the design intent and the code mostly aligns with it — single shell, unified settings panel, shared notice/repair primitives, parity on approval surfaces. A few real concerns flagged below, plus a CI flake worth re-running and one design tension worth confirming intent on.

### CI

**Go tests failed on `TestDetectVersionParsesOutputFromNonZeroExit`** in `internal/agentadapters/version_test.go:137`. This test isn't touched by this PR and passes 5/5 locally on master. Master's most recent Test run (`25909538840`) is green. Almost certainly a CI flake (the test creates a shell script that prints output then exits 1 — sensitive to runner-side process scheduling). Re-run before merge.

### Design / behavior change worth confirming

**Removal of \"Active external-agent usage\" panel from `CostsView`** (−61 lines). The rationale in code is: \"Only cross-chat Hecate-controlled provider calls belong here; active-chat adapter usage lives in ChatView where the reported values have context.\" That's RFC-aligned and the in-chat surface is the right place for adapter-reported usage that's only valid for one session. But operationally, this means **operators inspecting `/costs` will no longer see Codex/Claude-Code/Cursor spend at all** — they'd have to open the specific chat. PR description doesn't call this out. Confirm the intent matches what operators expect; if so, a one-line note in the description (or `docs/chat-sessions.md`) helps the next person who's looking for adapter spend on the wrong screen.

### Code-review concerns

1. **`HandleSetAgentChatSettings` non-atomic task+session update.** [internal/api/handler_agent_chat.go:479-498](internal/api/handler_agent_chat.go) updates `task.RTKEnabled` first, then `session.RTKEnabled`. If the second update fails (DB error mid-call), the task ends up with RTK on while the session stays off — divergent state, 500 to the caller. Matches the broader Hecate pattern (no cross-table transactions elsewhere either), so this is consistent rather than a regression, but worth a `// best-effort, no cross-table transaction` comment so the next reader doesn't try to \"fix\" it. Alternatively, transpose the order: update session first (it's the source-of-truth field that drives runtime behavior), then fan out to task.

2. **No test for the `agent_chat.runtime_mismatch` 409 path on settings.** `TestUpdateAgentChatSessionRenamesSession` covers rename; `TestUpdateAgentChatSessionRejectsEmptyTitle` covers validation. The `HandleSetAgentChatSettings` external-agent rejection (`renderAgentChatRuntimeKind(session) == \"external_agent\"`) is unguarded by tests. Recommend adding `TestSetAgentChatSettingsRejectsExternalAgent` mirroring the existing pattern — single-line copy of the rename test plus a runtime kind switch.

3. **Approval modal carries state across chat-target switches.** `ChatView` keeps `approvalModalID` outside the `chatTarget`-keyed state. Switching from an external-agent chat with a pending approval modal to a Hecate chat doesn't clear the modal; the next approval action would target the wrong runtime. Not a security issue (the resolve API rejects mismatched IDs), but a UX rough edge. Either gate the modal by `chatTarget` membership or clear `approvalModalID` in the existing `setChatTarget` effect.

4. **New telemetry attributes are stable public contract.** `internal/telemetry/semconv.go` adds three attrs:
   - `hecate.sandbox.rtk.enabled`
   - `hecate.sandbox.rtk.command.before`
   - `hecate.sandbox.rtk.command.after`

   `docs/telemetry.md` is updated (+9/-1) which is great. The before/after command strings are operator command text — confirm the RTK transformation never leaks secrets (env-var-bearing argv) into spans. If `command.before` ever contains substituted credentials, that's a structured PII leak into trace exporters. A redaction test against a known-secret-bearing input would lock that down.

5. **`pkg/types/task.go` `RTKEnabled` field is properly additive** (JSON-tagged, additive in serialization). Positional reflection isn't used anywhere in the repo I can see, and Tauri/SDK consumers go through JSON. No concern despite the visible field reorder.

6. **`sandbox.RTKAvailable()` exported and used in admin stats but unexercised by tests.** The check is just `exec.LookPath(\"rtk\")` — almost certainly correct — but operators relying on `/system/stats` to discover RTK availability will quietly see `false` on machines where `rtk` is installed under a non-PATH location (rare but possible). One sentence in `docs/agent-runtime.md` next to the new stats field would close the loop.

### Things that look good

- **RFC + code alignment on the unified shell.** `ChatView.tsx`'s `isHecateChat` / `isExternalAgentChat` gates funnel both runtimes through the same sidebar/settings/composer surface, exactly as the RFC scopes. `ChatNotice.tsx`'s `ChatNoticeFrame` primitives are the right shape for the \"shared notice tone\" goal.
- **`useRuntimeConsole` hook extraction** (+140/-17) is a real architectural win — pulls the runtime-aware state machinery out of the giant component into a testable unit.
- **`ChatAgentControls.tsx` consolidation** brings the per-agent picker / mode toggle / config controls into one file with the right separation between runtime-shared (`NewChatAgentButton`, `ChatAgentPicker`) and runtime-specific (`HecateToolsToggle`, `ExternalAgentConfigControls`) primitives.
- **Sqlite auto-migration for `rtk_enabled INTEGER NOT NULL DEFAULT 0`** in `internal/agentchat/sqlite.go` is the right shape — backward-compatible, no operator action required on upgrade.
- **`docs/runtime-api.md` (+68)** properly documents both new PATCH endpoints with request/response examples.
- **PR description \"Validation\" lists `go test`, `go vet`, `go test -tags e2e`, `bun typecheck/test/test:e2e`** — every relevant suite, matches what CI ran here (except the flake).

LGTM-modulo nits. The four code concerns above are all minor and addressable in follow-up commits without blocking. The CostsView behavior change is the only thing where I'd want explicit confirmation that the operator-facing flow matches your intent before merge.

PR #103, opened 2026-05-15
ci: gate desktop packaging behind tests

chicoxyzzy:
## Summary

- Fix PR Tauri validation so unsigned PR bundles are build-checked but not uploaded as workflow artifacts.
- Move PR desktop packaging into the main Test workflow behind the cheaper Go, TypeScript, e2e, Docker smoke, and Tauri Rust checks.
- Keep `tauri-build.yml` as a manual-only desktop rebuild/debug workflow.
- Document the CI workflow split in development, desktop, release, and Tauri agent docs.

## Verification

- `ruby` YAML parse for `test.yml`, `tauri-build.yml`, `_tauri-shared.yml`, and `release.yml`
- `actionlint .github/workflows/test.yml .github/workflows/tauri-build.yml .github/workflows/_tauri-shared.yml .github/workflows/release.yml`

## Notes

This keeps release behavior unchanged: tagged releases still upload signed/notarized desktop artifacts through `release.yml`.

### DIFF
--- a/.github/workflows/_tauri-shared.yml
+++ b/.github/workflows/_tauri-shared.yml
@@ -153,7 +153,9 @@ jobs:
           GITHUB_TOKEN: ${{ github.token }}
         with:
           projectPath: tauri
-          args: --bundles ${{ matrix.bundles }}
+          args: >-
+            --bundles ${{ matrix.bundles }}
+            --config '{"bundle":{"createUpdaterArtifacts":false}}'
 
       # Tagged release build. Uploads bundles to the existing GitHub Release and
       # signs/notarizes when the corresponding secrets are configured.
--- a/.github/workflows/_tauri-shared.yml
+++ b/.github/workflows/_tauri-shared.yml
@@ -134,14 +134,13 @@ jobs:
       - name: Stamp Tauri version
         run: bun scripts/stamp-version.ts
 
-      # Single tauri-action call serves both modes:
-      #   - tagName empty  → build only (PR validation via tauri-build.yml)
-      #   - tagName set    → upload to the existing GitHub Release for that tag
-      #                      (release.yml after a `v*` tag push)
-      #
-      # macOS signing/notarization protections are documented in the
-      # env: block below.
-      - name: Build Tauri bundles
+      # Keep PR builds and release builds as separate steps. Setting signing
+      # variables to an empty string is not the same as omitting them: Tauri's
+      # bundler still treats some present-but-empty vars as "signing requested"
+      # and fails while decoding/importing them. PR validation must pass no
+      # signing/notarization env at all.
+      - name: Build Tauri bundles (unsigned PR)
+        if: inputs.tagName == ''
         uses: tauri-apps/tauri-action@v0
         env:
           # `github.token` is the per-job-run token, available in every
@@ -151,6 +150,17 @@ jobs:
           # inherit` — see that file's job-level comment for why
           # withholding inherit is the load-bearing protection
           # against PR-side secret exfiltration.
+          GITHUB_TOKEN: ${{ github.token }}
+        with:
+          projectPath: tauri
+          args: --bundles ${{ matrix.bundles }}
+
+      # Tagged release build. Uploads bundles to the existing GitHub Release and
+      # signs/notarizes when the corresponding secrets are configured.
+      - name: Build Tauri bundles (signed release)
+        if: inputs.tagName != ''
+        uses: tauri-apps/tauri-action@v0
+        env:
           GITHUB_TOKEN: ${{ github.token }}
           # Apple signing/notarization secrets. Two independent
           # protections in series:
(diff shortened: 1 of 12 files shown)


chicoxyzzy (commented):
Solid change overall — gating the expensive matrix behind the cheap tests is the right shape, and the underlying fixes (split unsigned-PR / signed-release steps, `--config createUpdaterArtifacts:false`, `goexe` suffix, `updater:default` capability) are exactly the same root causes I hit on #101 and that #102 also addresses. Glad to see this landing in the workflow proper rather than as a per-PR cherry-pick.

A few observations / questions, none blocking:

**Strengths**

- `--config '{"bundle":{"createUpdaterArtifacts":false}}'` on the unsigned PR step is the precise fix for the Linux "incorrect updater private key password: Missing comment in secret key" error I was diagnosing yesterday. Cleaner than stripping env vars — it tells the bundler not to even try the signing path.
- Job-level `permissions: contents: write, actions: write` on `tauri-desktop` correctly handles the GitHub permission-cap rule that takes effect at workflow startup, before any `if:` gates evaluate. The active CI run on this PR confirms the matrix actually starts (no startup_failure) — the gating works as written.
- Resource discipline is good: the matrix's `if:` waits for every cheap job to be `success` OR `skipped`, so path-filtered tests don't accidentally block the desktop build, and a real failure on any of them short-circuits.
- `tauri-build.yml` shrinking to manual-only with `concurrency: cancel-in-progress: true` retained makes "rerun for debug" idempotent in the way it should be.
- Doc updates land in all the right places (`docs/development.md` CI section, `docs/desktop-app.md`, `docs-ai/skills/tauri/SKILL.md`, `docs-ai/tasks/release.md`). The flow chart implicit in the workflow table is much easier to follow now.

**Questions / minor nits**

1. **The `needs:` and `if:` blocks on `tauri-desktop` enumerate every cheap job by name.** Adding a future top-level job (say `test-something-new`) requires remembering to thread it through both lists, or the desktop matrix starts without waiting for it. Worth a brief comment in the YAML acknowledging that constraint, or — bigger refactor — funnel everything through a single \"all required passed\" gate job. Not blocking.

2. **`cache: true` removed from three `actions/setup-go@v6` invocations.** Unrelated to gating, unexplained, and the action's default has been `cache: true` since v3 so this is semantically a no-op — but it reads as a deliberate change in the diff. If there's a reason (cache contention with `Swatinem/rust-cache`?) worth a one-line comment. If accidental, drop the hunk.

3. **`mapfile -t files < <(git ls-files '*.go')`** is a real robustness improvement (handles filenames with spaces / newlines) but is also unrelated to the gating goal. Easier review history if it's a separate commit, but not worth blocking on.

4. **No `tauri-desktop` workflow-level `paths-ignore` for `website/**`** — the existing `paths-ignore` on `test.yml` already excludes it, and `changes.desktop` filter only matches relevant inputs, so the matrix won't fire on website-only changes. Good. Worth noting in a quick comment so the next reader doesn't worry about it.

5. **PR validation no longer uploads bundles as artifacts.** Reviewers who want to test-launch a built bundle pre-merge now need to manually dispatch `tauri-build.yml`. `docs/desktop-app.md` and `docs/development.md` describe the new flow but neither calls out this workflow change for reviewers explicitly. Consider adding a one-liner under \"PR validation\" in `docs/desktop-app.md`: \"To test-launch a bundle pre-merge, dispatch `tauri-build.yml` manually from the Actions tab.\"

6. **`TAURI_SIGNING_PRIVATE_KEY: ${{ secrets.TAURI_UPDATER_PRIVATE_KEY }}`** is now unconditional in the signed-release step, which is correct since the step itself is gated by `if: inputs.tagName != ''`. The old `inputs.tagName != '' && ...` inside the expression was a redundant double-gate; removing it is a clean simplification.

7. **`capabilities.json` now includes `updater:default`.** Why this wasn't needed before: previous runs all `startup_failure`'d before the build ever reached the bundler, so the missing capability never surfaced. Adding it is correct and matches `bundle.createUpdaterArtifacts: "v1Compatible"` in `tauri.conf.json`.

**Verification — what's running right now**

CI on this PR currently has all 11 cheap jobs green and the 3 Tauri desktop matrix legs `in_progress` — the actual end-state of the matrix legs is the real proof. If they succeed, the gating shape AND the signing/notarization splits are jointly validated against macOS, Linux, and Windows runners for the first time since `35b59cd4` (2026-05-01).

LGTM modulo the small nits above. Approving as a comment-review since I'm not formally a CodeOwner here.

chicoxyzzy:
Addressed the useful follow-up nits in `f2aa767e`:

- added a YAML comment explaining that `tauri-desktop.needs` is intentionally explicit and must stay in sync with the cheap/required PR checks;
- documented that reviewers who need a pre-merge desktop bundle should manually dispatch `tauri-build.yml` from the PR branch;
- refreshed the desktop/release/signing docs that still implied `tauri-build.yml` owned PR validation.

Small correction on the cache note: the unsupported `cache: true` input was removed from `oven-sh/setup-bun@v2`, not from `actions/setup-go@v6`; Go caching remains enabled.

chicoxyzzy (commented):
All my nits addressed cleanly — thanks for the quick turnaround:

| Concern | Resolution |
|---|---|
| `needs:` list brittleness | `f2aa767e` adds an inline comment in `test.yml`: *"Keep this explicit list in sync with every cheap/required PR check. The desktop matrix is intentionally last so a failing unit/e2e/smoke check does not burn macOS/Windows/Linux packaging minutes."* Makes the brittleness intentional and documented. |
| Unexplained `cache` removal | `531e72ac` split out as its own commit (\"ci: remove unsupported Bun cache input\") with a real explanation — `oven-sh/setup-bun@v2` now warns that `cache` is not a valid input. My review misread the diff as `setup-go`; it was `setup-bun`. Clean separation now. |
| Reviewer-facing doc for manual dispatch | Added in 5 places: `docs/desktop-app.md`, `docs/development.md`, `docs-ai/skills/tauri/SKILL.md`, `docs-ai/tasks/release.md`, plus refreshed `docs/desktop-updater-signing.md`, `docs/macos-signing.md`, `docs/release.md` so every signing/release doc names the new shape. |
| `_tauri-shared.yml` header comment | Now lists all three callers (`release.yml`, `test.yml`, `tauri-build.yml`) with their roles. |
| `uploadArtifacts` input description | Tweaked to reflect new semantics: *\"Upload built bundles as workflow artifacts for manual/debug callers.\"* |

The earlier CI run on this branch (`25908766333`, since cancelled when the new commits landed) showed all three desktop matrix legs completing `success` — macOS / Windows / Linux — for the first time since `35b59cd4`. The fix is jointly validated against signing/notarization splits AND the gating shape. The current re-run (`25909372775`) is in flight on the latest commits.

Nothing left from my side. LGTM — approve.

PR #104, opened 2026-05-15
docs(rfcs): add Embeddings RFC — POST /v1/embeddings end-to-end

chicoxyzzy:
## Summary

Adds a design-notes RFC at `docs/rfcs/embeddings.md` for a provider-routed `POST /v1/embeddings` endpoint.

Hecate routes chat traffic but doesn't expose an embeddings endpoint. Operators wiring up retrieval (Hecate Chat memory, RAG, external brains like gbrain) currently bypass the gateway entirely — secrets, usage, and cost data all leak around it. The RFC scopes the smallest framework that closes the gap.

## What the RFC commits to

- `POST /v1/embeddings` with the OpenAI request/response shape verbatim — single string or array input, optional `dimensions` / `encoding_format`, OpenAI-shaped envelope back.
- New optional `Embedder` interface on the provider adapter, parallel to `Streamer`. Adapters that don't implement it surface as `capability: chat` only; the router returns a stable `400 model_unsupported_capability` for embedding requests against their models.
- Provider coverage v1: OpenAI, Voyage (new adapter — Anthropic's recommended embeddings vendor for Claude RAG), Gemini, Azure OpenAI, and the bundled `llama-server`. Anthropic / DeepSeek / Groq / Perplexity stay opt-out — no upstream embeddings API.
- `internal/modelcaps/modelcaps.go:76` flips from "drop embedding-named models" to "tag them with `capability: embedding`." `/v1/models?capability=…` grows an optional filter; default stays `chat` for backwards compatibility.
- `internal/llamacpp/runtime_process.go` appends `--embeddings` when a per-`InstalledModel` capability bit is set; curated catalog entries for `nomic-embed-text` / `bge-small` ship with the bit on.
- Same `Usage` / `CostBreakdown` pipeline records embedding rows alongside chat; `pricebook.json` grows `embeddings_per_million_usd`.

## Explicit non-goals for v1

- **Failover** — would silently corrupt vector stores by indexing into a different model's space.
- **Gateway-level caching** — content-addressed cache is a real win but needs its own RFC.
- **Reranking** (`/v1/rerank`) — different endpoint, only Voyage and Cohere ship it.
- **Multimodal embeddings** — wire shape similar, deferred until text proves out.
- **Streaming** — embeddings APIs return single JSON envelopes.

## Phased landing

| Phase | Scope |
|---|---|
| v1.0 | Adapter interface, OpenAI adapter, `/v1/embeddings` route, capability filter, basic tests |
| v1.1 | Voyage adapter (new file), Gemini adapter (extend existing) |
| v1.2 | Local-model integration: per-model `embeddings` capability bit, runtime args toggle, curated catalog entries |
| v1.3 | Cost-pricing, dashboard surface, operator-facing `/hecate/v1/providers/{id}/embedding-models` listing |

Each phase is independently mergeable; v1.0 is the contract everything downstream depends on.

## Test plan

- [ ] Reviewers agree the OpenAI wire shape is the right contract for v1
- [ ] Reviewers agree Voyage justifies a new adapter (vs extending OpenAI) — Voyage's `input_type` field has retrieval-quality implications worth surfacing intentionally
- [ ] Reviewers agree the default `/v1/models?capability=chat` filter preserves backwards compatibility for tools that walk the model list
- [ ] Reviewers agree the v1 scope holds — failover and caching out, single-provider routing in
- [ ] No code lands until this RFC is accepted

Index entry added at [docs/rfcs/README.md](docs/rfcs/README.md) as "Embeddings — Not implemented."


chicoxyzzy:
All five points addressed in [`6df21e63`](https://github.com/hecatehq/hecate/pull/104/commits) (rewrite touches ~240 lines):

| # | Concern | Resolution |
|---|---|---|
| P1 | Pinned routing / no fallback contract | New **Routing semantics** section codifies exact-match resolution, no default-model fallback (unknown model → `404 model_not_configured`), no health-based provider failover (degraded primary → upstream error, never a substitute). Lives in a separate `routeEmbedding` function rather than a flag on the chat router so a future chat-side change can't regress to embedding failover. New acceptance criterion locks the behavior. |
| P2 | Stale usage/cost architecture | Dropped `internal/usage`, `internal/cost`, `pricebook.json`, "Cost dashboard." Rewritten to match what's actually there: per-event `governor.UsageEvent` rows with `CostMicros` computed inline by each adapter, surfaced through the Usage view. Verified against `TestUsageMutationEndpointsAreRemoved`. |
| P2 | `modelcaps` mischaracterization | Corrected — `modelcaps.go` does NOT drop embedding models, it sets `tool_calling: "none"`. And `/v1/models` exposes capabilities under `metadata.capabilities`, not top-level singular `capability`. Now proposes a new `ModelCapabilities.Kind` field (`"chat"` / `"embedding"`), additive and omitempty, surfaced as `metadata.capabilities.kind`. |
| P2 | `encoding_format=base64` inconsistency | Deferred to v1.1. v1.0 ships `float` only; internal `EmbeddingResponse.Embeddings` stays `[][]float32` with no encoding-format field, eliminating the round-trip waste and the open question. Phasing updated; v1.1 introduces the preserve-upstream-bytes path either via an `EmbeddingVector` wrapper or a split adapter method. |
| P3 | LiteLLM mention in gbrain note | `OPENAI_BASE_URL / LITELLM_BASE_URL` → `the OpenAI-compatible base URL (OPENAI_BASE_URL)`. |

Also tightened the Risks section so it doesn't say "failover off by default" — it's permanently off, and the future-phasing line now reads `Failover stays permanently off for embeddings; it is NOT a future phase.`

RFC index entry refreshed to reflect the new shape (separate router, pinned routing, `ModelCapabilities.Kind`, `float` only in v1.0).

PR #107, opened 2026-05-15
refactor(workspace,acp): introduce Workspace abstraction + editor-owned mode negotiation

chicoxyzzy:
## Summary

Refactors the orchestrator's executor boundary into a `workspace.Workspace`
interface, adds an editor-owned ACP implementation behind it, and wires
`HECATE_WORKSPACE_MODE` capability negotiation into the bridge's `initialize`
handshake. This is the prep work for editor-owned ACP sessions where Zed,
JetBrains, or another ACP host owns file writes and terminal execution via
reverse-RPC.

**This PR negotiates the wire and ships both workspace implementations. It
does not yet route reverse-RPC end-to-end** — the gateway-side orchestrator
still uses `LocalWorkspace` in every mode. The transport piece (bridge↔gateway
duplex IPC so the orchestrator can reach back through the bridge to the
editor) is the subject of a follow-up RFC.

### What changed

- New `internal/workspace` package with a `Workspace` interface, a
  `LocalWorkspace` wrapping the existing sandbox, and a new `ACPWorkspace`
  that dispatches reverse-RPC (`fs/read_text_file`, `fs/write_text_file`,
  `terminal/create`, `terminal/output`, `terminal/wait_for_exit`,
  `terminal/kill`, `terminal/release`, `session/request_permission`).
- New `Terminal` + `Permitter` interfaces — long-lived terminals with a
  streaming chunk channel, plus an optional permission gate that defers to
  the editor.
- Orchestrator dependency swap: `sandbox.Executor` → `workspace.Workspace`
  in three struct fields and four function signatures. No behavior change in
  the local path; the abstraction is the only delta.
- New `acp.Dispatcher.Call(ctx, method, params)` generic outbound reverse-RPC
  with per-call channel routing keyed by id. Separate id namespace from the
  permission flow so the two response paths stay decoupled.
- `acp.ResolveWorkspaceMode` + granular `FSCapability` fields
  (`readTextFile`, `writeTextFile`). Default flips from `hecate-owned` to
  `auto`. `editor-owned` mode fails `initialize` loudly when the editor lacks
  the required capabilities (`fs.readTextFile`, `fs.writeTextFile`,
  `terminal`) rather than silently downgrading.
- `docs/acp.md` updated: configuration table, status section flipped from
  "not implemented" to "wired but transport pending."

### What's deliberately out of scope

- The bridge↔gateway transport for reverse-RPC. Today the gateway-side
  orchestrator always uses `LocalWorkspace`; `ACPWorkspace` exists but isn't
  reachable from the orchestrator yet. A follow-up RFC will pick between
  WebSocket-on-existing-HTTP, UDS, and inverting the process layout so the
  dispatcher lives in the gateway and the bridge becomes a stdio relay.
- ACP auth (`env_var` method per the ACP auth-methods RFD). Deferred until
  the transport question is settled.

## Verification

- `go test ./... -race -count=1` — 1803 passed in 37 packages
- New `internal/acp/workspace_mode_test.go` covers the resolver matrix
  (auto fallback, editor-owned guard, case-insensitive parsing, invalid mode
  rejection)
- New dispatcher tests for the editor-owned-without-caps rejection, auto
  fallback to hecate-owned, and auto promotion to editor-owned with full caps

### DIFF
--- a/cmd/hecate-acp/main.go
+++ b/cmd/hecate-acp/main.go
@@ -105,7 +105,7 @@ func configFromEnv() (bridgeConfig, error) {
 		GatewayURL:    firstNonEmpty(gatewayURL, defaultGatewayURL),
 		AgentName:     firstNonEmpty(os.Getenv("HECATE_AGENT_NAME"), "Hecate"),
 		AgentVersion:  version.Version,
-		WorkspaceMode: firstNonEmpty(os.Getenv("HECATE_WORKSPACE_MODE"), "hecate-owned"),
+		WorkspaceMode: firstNonEmpty(os.Getenv("HECATE_WORKSPACE_MODE"), acp.WorkspaceModeAuto),
 		ApprovalRoute: firstNonEmpty(os.Getenv("HECATE_APPROVAL_ROUTE"), "editor"),
 		OTel:          bridgeOTelFromEnv(),
 	}
--- a/docs/acp.md
+++ b/docs/acp.md
@@ -53,11 +53,14 @@ Implemented:
 
 Not implemented yet:
 
-- Editor-owned workspace reverse-RPC calls. Hecate-owned mode is the current
-  implementation: work runs through the gateway task runtime. The future
-  editor-owned mode would let an ACP host such as Zed or JetBrains own file and
-  terminal operations through ACP reverse-RPC, including terminal lifecycle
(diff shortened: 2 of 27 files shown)


chicoxyzzy:
@copilot asking for round 2. All fixed

PR #112, opened 2026-05-16
refactor(ui): consolidate localStorage sync via usePersistedState

chicoxyzzy:
## Summary

Phase 3 of the UI cleanup plan ([`graceful-giggling-spring`](https://github.com/hecatehq/hecate/tree/master)). One hook replaces ten inline `useEffect(setItem, [val])` sync sites in `useRuntimeConsole` and the two bespoke read/write helper pairs (queued chat messages, chat-target map), plus the theme + workspace persistence in `App` / `AppShell`.

### What's new

`ui/src/lib/persistedState.ts`:

```ts
const [model, setModel] = usePersistedState("hecate.model", parseStoredString, "", {
  shouldRemove: v => v === "",
});

const [queuedMessages, setQueuedMessages] = usePersistedState(
  "hecate.queuedChatMessages",
  parseStoredJSON(parseQueuedChatMessageList),
  [],
  { shouldRemove: v => v.length === 0 },
);
```

The `parse(raw)` callback returns `T | null`. **null wipes the key + logs via `lib/log.ts`**, then falls back. This is the "loud failure" contract.

### Migrated (11 keys)

| Key | File | Notes |
|---|---|---|
| `hecate.workspace.v2` | `App.tsx` | parse via `VALID_WORKSPACE_IDS` Set |
| `hecate.theme` | `AppShell.tsx` | `shouldRemove: v === "system"` |
| `hecate.model` | `useRuntimeConsole.ts` | shouldRemove on empty |
| `hecate.systemPrompt` | `useRuntimeConsole.ts` | string passthrough |
| `hecate.providerFilter` | `useRuntimeConsole.ts` | cast to ProviderFilter |
| `hecate.chatTarget` | `useRuntimeConsole.ts` | normalizeStoredChatTarget |
| `hecate.agentAdapterID` | `useRuntimeConsole.ts` | string passthrough |
| `hecate.agentWorkspace` | `useRuntimeConsole.ts` | string passthrough |
| `hecate.chatSessionID` | `useRuntimeConsole.ts` | shouldRemove on empty |
| `hecate.agentChatSessionID` | `useRuntimeConsole.ts` | shouldRemove on empty |
| `hecate.chatTargetBySessionID` | `useRuntimeConsole.ts` | parse + custom serialize for Map |
| `hecate.queuedChatMessages` | `useRuntimeConsole.ts` | per-item filter, shouldRemove on empty |

### Retired

- `readLocalStorage`, `readStoredQueuedChatMessages`, `writeStoredQueuedChatMessages`, `readStoredChatTargetsBySessionID`
- Mount-time read useEffect (8 `readLocalStorage` reads)
- 9 inline write useEffects
- 200 ms queued-messages debounce + pagehide flush (the new hook writes synchronously so the flush is redundant)

### Semantics change to call out

The queued-chat-messages defensive `typeof item.foo === "string" ? item.foo : ""` guards previously massaged corrupt records into half-zeroed objects, which then re-persisted on the next mutation and outlived their original version. The new contract:

1. Per-item filtering preserved (one corrupt record drops just that record, not the whole queue).
2. **Array-shape mismatch (parse returns null) wipes the key** via `usePersistedState` and logs through `lib/log.ts`.

This is the "remove hiding legacy" item from the plan and is intentional. Existing valid stored data is unaffected.

## Test plan

- [x] `bun run typecheck` — clean (`tsgo -b`)
- [x] `bunx vitest run` — 765/765 pass (was 752; +13 new `persistedState.test.ts`)
- [x] `useRuntimeConsole.test.tsx` (96 KB) is the regression net for the 11 storage keys and runs unchanged
- [x] Theme-bootstrap inline script in `index.html` continues to win the race against React mount (theme key contract preserved)
- [x] `useDesktopUpdate`'s sessionStorage usage is untouched (its own concern)

### DIFF
--- a/ui/src/app/App.tsx
+++ b/ui/src/app/App.tsx
@@ -1,24 +1,29 @@
-import { useEffect, useLayoutEffect, useState } from "react";
+import { useEffect, useLayoutEffect } from "react";
 
 import { ConsoleShell, getAvailableWorkspaces, type WorkspaceID } from "./AppShell";
 import { useRuntimeConsole } from "./useRuntimeConsole";
+import { usePersistedState } from "../lib/persistedState";
 import { isTauriRuntime } from "../lib/tauri";
 
 const WORKSPACE_STORAGE_KEY = "hecate.workspace.v2";
 
+const VALID_WORKSPACE_IDS = new Set<WorkspaceID>(["overview", "runs", "chats", "connections", "usage", "settings"]);
+const parseWorkspaceID = (raw: string): WorkspaceID | null =>
+  VALID_WORKSPACE_IDS.has(raw as WorkspaceID) ? (raw as WorkspaceID) : null;
+
 export default function App() {
   const { state, actions } = useRuntimeConsole();
-  const [preferredWorkspace, setPreferredWorkspace] = useState<WorkspaceID>(() => {
-    const saved = localStorage.getItem(WORKSPACE_STORAGE_KEY);
-    return (saved as WorkspaceID) ?? "chats";
-  });
+  const [preferredWorkspace, setPreferredWorkspace] = usePersistedState<WorkspaceID>(
+    WORKSPACE_STORAGE_KEY,
+    parseWorkspaceID,
+    "chats",
+  );
 
   const workspaces = getAvailableWorkspaces();
   const activeWorkspace: WorkspaceID =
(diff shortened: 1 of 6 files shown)


chicoxyzzy:
@copilot review

PR #134, opened 2026-05-17
refactor(ui): drop the model-chat transcript placeholder + duplicate spec [skip desktop]

chicoxyzzy:
## Summary

Two loose ends that surfaced during the model-chat backend removal but were too small to land alongside the deletion. Both deletions, no behavior change.

```
3 files changed, 19 insertions(+), 93 deletions(-)
```

## What changes

`ChatTranscript` still carried two relics from the legacy model-chat data path that lost its callsite when `/chat/sessions` REST went away:

1. **`callsByID: Map<string, LegacyProviderCallInfo>`** — every caller passed an empty map; every `call?.x` consumer evaluated to undefined. The whole `produced_by_call_id` lookup was dead.
2. **`isAgentChat: boolean`** — derived in `ChatView` as `isHecateChat || chatTarget === "external_agent"`, which today always reduces to true (the runtime_kind enum has no "no chat" value). Every `isAgentChat && X` ternary always took the X branch; the `!isAgentChat` streaming pane never rendered.

Removed:
- `LegacyProviderCallInfo` type and the `callsByID` prop on `ChatTranscript`.
- `isAgentChat` prop on `ChatTranscript`; the now-unconditional branches inlined.
- The `!isAgentChat`-only streaming pane (~25 lines of JSX) and its branding/model props (`selectedAgent`, `selectedConfiguredProvider`, `selectedRuntimeProvider`, `hecateChatModelValue`) + the `BrandAvatar` import they used.
- The `messageBrand(message, provider, isAgentChat, isHecateAgentChat)` signature drops both `provider` (always undefined from the dead call lookup) and `isAgentChat` (always true).
- The empty `callsByID = new Map<string, never>()` setup in `ChatView` and the prop pass it fed.

Also drops one duplicate e2e test:
- `ui/e2e/settings.spec.ts` had `"Usage workspace shows the empty usage state"` — this test moved out to `ui/e2e/usage.spec.ts` in [#131](https://github.com/hecatehq/hecate/pull/131); the duplicate stuck around as an unstaged-file mistake.

## What stays

`TranscriptMessageRow`'s `promptTokens` / `completionTokens` / `costUsd` props stay on the component — its own tests still exercise them, and they remain part of the row's public contract if a future caller has the data.

## Test plan

- [x] `cd ui && bun run typecheck` — clean (`tsgo -b`)
- [x] `cd ui && bunx vitest run` — 939 / 939 pass (40 files)
- [x] `cd ui && bun run test:e2e` — 60 / 60 pass (was 61, the duplicate is gone)

## Why `[skip desktop]`

Pure TSX cleanup under `ui/src/features/chats/` + a one-test deletion in `ui/e2e/`. No `tauri/**` or `desktop_force` paths touched.

### DIFF
--- a/ui/e2e/settings.spec.ts
+++ b/ui/e2e/settings.spec.ts
@@ -49,8 +49,3 @@ test("maintenance 'Clean up now' fires POST request", async ({ page }) => {
   await page.getByRole("button", { name: /Clean up now/i }).click();
   await expect.poll(() => posted).toBe(true);
 });
-
-test("Usage workspace shows the empty usage state", async ({ page }) => {
-  await page.locator(".hecate-activitybar [aria-label^='Usage']").click();
-  await expect(page.locator("text=No cloud usage recorded yet")).toBeVisible();
-});
--- a/ui/src/features/chats/ChatTranscript.tsx
+++ b/ui/src/features/chats/ChatTranscript.tsx
@@ -2,20 +2,8 @@ import { useEffect, useRef, useState, type ReactNode } from "react";
 
 import { useRuntimeConsoleContext } from "../../app/RuntimeConsoleContext";
 import { formatDurationMs } from "../../lib/format";
-import type { AgentAdapterRecord, ChatActivityRecord, ChatSegmentRecord, ChatTimingRecord, ChatUsageRecord, ConfiguredProviderRecord, ProviderRecord } from "../../types/runtime";
-
-// Placeholder shape kept for the produced_by_call_id lookup in the transcript.
-// The legacy /chat/sessions data path used to populate this map with rich
-// per-call records; agent-chat messages don't, so callers pass an empty map.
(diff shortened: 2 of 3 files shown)


chicoxyzzy (commented):
Reviewed the current head (`94cb13c`). I don’t see any blocking issues.

The cleanup matches the current architecture: the old `callsByID` map had no population path after the legacy chat-state removal, and the deleted non-agent streaming pane was behind a branch that no longer represents a live transcript mode. Current chat rows still render running assistant placeholders from chat session snapshots.

Checks run locally on the PR head:

```sh
cd ui && bun run typecheck
cd ui && bunx vitest run src/features/chats/ChatView.test.tsx src/features/transcript/TranscriptMessageRow.test.tsx
```

Both passed. The duplicate Usage e2e removal also looks correct because `ui/e2e/usage.spec.ts` still covers the empty usage state.

PR #137, opened 2026-05-17
refactor(providers): centralize protocol dispatch in one table [skip desktop]

chicoxyzzy:
## Summary

Two sites in `internal/providers/runtime_manager.go` branched on the provider protocol via inline string compares:

- **line 231**: gated the gateway-wide `AnthropicCacheDisabled` toggle on `strings.EqualFold(cfg.Protocol, \"anthropic\")`.
- **lines 385-390**: a `switch strings.ToLower(strings.TrimSpace(providerCfg.Protocol))` chose the Provider constructor (Anthropic vs. default OpenAI-compatible).

Adding a third protocol meant editing both — and drift between the constructor switch and the cache-gating check was a latent bug. Move the per-protocol facts to one table.

```
2 files changed, 64 insertions(+), 9 deletions(-)
```

## New: `internal/providers/protocol_dispatch.go`

```go
type protocolDispatch struct {
    Constructor            func(cfg, logger) Provider
    SupportsAnthropicCache bool
}

var protocolDispatchByName = map[string]protocolDispatch{
    "anthropic": { Constructor: ..., SupportsAnthropicCache: true },
    "openai":    { Constructor: ..., SupportsAnthropicCache: false },
}

// Unknown protocols fall back to the OpenAI-compatible entry,
// preserving buildProviders' prior default semantics. The bool
// signals whether the protocol was explicitly registered.
func lookupProtocolDispatch(protocol string) (protocolDispatch, bool)
```

## Edit: `runtime_manager.go`

- `buildProviders`' switch becomes one lookup + `entry.Constructor(...)` call.
- The cache-marker gating becomes `entry.SupportsAnthropicCache`, so it can't drift from the constructor's actual capability.

A third protocol (Google, Bedrock, etc.) now registers one entry in the dispatch map; both sites pick it up automatically.

## Test plan

- [x] `go vet ./...` — clean
- [x] `gofmt -s -l .` — clean

## Why `[skip desktop]`

Pure Go-side refactor under `internal/providers/`. No `ui/**` or `tauri/**` changes.

### DIFF
--- a/internal/providers/protocol_dispatch.go
+++ b/internal/providers/protocol_dispatch.go
@@ -1,0 +1,60 @@
+package providers
+
+import (
+	"log/slog"
+	"strings"
+
+	"github.com/hecate/agent-runtime/internal/config"
+)
+
+// protocolDispatch describes one inbound-protocol family the gateway can
+// instantiate. It centralizes the per-protocol facts that callers used
+// to discover via scattered `Protocol == "anthropic"` checks:
+//
+//   - Constructor — how to build a Provider for this protocol.
+//   - SupportsAnthropicCache — whether the gateway-wide Anthropic prompt
+//     cache toggle (config.Providers.AnthropicCacheDisabled) is
+//     meaningful for instances on this protocol. Only the Anthropic
+//     adapter respects the flag today; OpenAI-compatible providers
+//     ignore it because the upstream has no equivalent knob.
+//
+// Add a third protocol by registering one entry below. The two original
+// dispatch sites (buildProviders constructor switch and the cache-marker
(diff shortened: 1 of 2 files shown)


chicoxyzzy (commented):
No blocking findings. I checked the protocol dispatch refactor and verified the important compatibility points: unknown/empty protocols still fall back to the OpenAI-compatible provider, and Anthropic cache disabling remains gated only for the Anthropic protocol entry.

Checks run:

- `gofmt -l internal/providers/protocol_dispatch.go internal/providers/runtime_manager.go`
- `git diff --check origin/master...HEAD`
- `go test ./internal/providers`

The table shape makes future protocol additions cleaner without changing current runtime behavior.

PR #141, opened 2026-05-17
refactor(api): extract task stream + events handlers [skip desktop]

chicoxyzzy:
## Summary

Finishes the per-concern split of `handler_tasks.go`. The SSE stream handler, the event-append handler, the snapshot builder, and the shared task/run loaders + SSE plumbing move to a sibling so each handler file owns one concern.

```
handler_tasks.go:        1613 → 1147 LOC
handler_tasks_stream.go: new, 494 LOC
```

## What moves to `handler_tasks_stream.go`

**Stream + events (3 handlers)**
- `HandleTaskRunStream`
- `HandleTaskRunEvents`
- `HandleAppendTaskRunEvent`

**Snapshot/event decoders + helpers**
- `buildTaskRunStreamState`
- `decodeTaskRunEventData`
- `decodeTurnCostFromEventData`
- `parseAfterSequence`

**Shared loaders + SSE plumbing**
- `loadAuthorizedTask`
- `loadAuthorizedTaskRun`
- `writeSSEHeaders`

The loaders are used by other handler files too (lifecycle, patches); they're file-scoped to `handler_tasks_stream.go` but visible to the whole package since everything is `package api`. Bundling them with the stream handler clusters the "read a task/run with auth checks" logic next to the file that uses them most.

## What stays in `handler_tasks.go`

CRUD (`HandleCreateTask`, `HandleTasks`, `HandleTask`, `HandleDeleteTask`, `HandleStartTask`), runs/steps (`HandleTaskRuns`, `HandleTaskRun`, `HandleTaskRunSteps`, `HandleTaskRunStep`), and the `buildTaskItem` / `buildTaskActivityItems` / `sortTaskActivityItems` / `cloneActivitySummary` helpers used by both CRUD and stream paths.

## What doesn't change

- The receiver type stays `*Handler` everywhere; no public API change.
- Mux registrations in `server.go` are unchanged.
- Test files unchanged.
- Imports per file pruned via `goimports`.

## Test plan

- [x] `go vet ./...` — clean
- [x] `gofmt -s -l .` — clean

## Why `[skip desktop]`

Pure Go-side file split under `internal/api/`. No `ui/**` or `tauri/**` changes.


chicoxyzzy:
Reviewed #141. No blocking findings; `go test ./internal/api` passes in the PR worktree.

Non-blocking note: the extraction copied `HandleRetryTaskRunFromTurn`'s doc comment into `internal/api/handler_tasks_stream.go` around line 261, where it now sits detached above `parseAfterSequence`. The real handler still has its comment in `handler_tasks.go`, so this is just stale source-comment noise from the split.

chicoxyzzy:
Addressed the stale-comment nit. The orphaned `HandleRetryTaskRunFromTurn` docblock that ended up above `parseAfterSequence` in handler_tasks_stream.go is gone (8 lines).

PR #151, opened 2026-05-17
feat(api): add http.server.request OTel span at handler entry [skip desktop]

chicoxyzzy:
## Summary

Hecate handlers already create per-subsystem spans via the in-memory profiler (router decision, provider call, governor check, etc.) and those spans propagate to the OTLP exporter pipeline. What was missing: the top-level `http.server.request` envelope they should hang off. The exporter saw the children but never the parent, so dashboards that filter on `http.server.request` would never light up — and traces in the operator UI's `/hecate/v1/traces` view had nothing tying them to the request route.

Add a thin middleware that opens one OTel span per inbound HTTP request and wire it into the Chain immediately after the existing request-id middleware.

```
3 files changed, 124 insertions(+)
```

## What changes

**`internal/api/middleware.go`**

- New `OTelHTTPSpanMiddleware` decorator. Opens a span named `"http.server.request"` on the package tracer (`"github.com/hecate/agent-runtime/internal/api"`), records OTel HTTP semconv attributes (`http.request.method`, `http.route`, `http.response.status_code`), and adds the operator-visible `hecate.request_id` attribute when one is already in context.
- The `http.route` attribute reads `r.Pattern` (Go 1.22+ mux pattern), populated by `http.ServeMux` during dispatch; we read it AFTER `next.ServeHTTP` so the value reflects the matched pattern. Falls back to `r.URL.Path` for paths that never went through a pattern-aware mux (e.g. notfound handler).
- 5xx responses set `span.Status = Error` so OTel-aware backends can surface them without re-deriving the threshold.

**`internal/api/server.go`**

- Chain ordering: `TraceContext → RequestID → OTelHTTPSpan → SameOrigin → Logging → Recovery`. The order matters: `TraceContext` populates the parent span context, `RequestID` populates the `request_id` attribute, and `OTelHTTPSpan` opens the span that every downstream subsystem span becomes a child of through the request context.

**`internal/api/middleware_test.go`**

- `TestOTelHTTPSpanMiddlewareEmitsRequestSpan` stands up an in-memory tracer + recorder, drives a request through the middleware, and asserts span name, attributes, status code, and the `hecate.request_id` correlation.

## What doesn't change

- In-memory profiler stays intact. Operators load `/hecate/v1/traces` from it; deleting that view is an operator-visible regression. OTel is the parallel exporter surface, not a replacement.
- No exported symbol moves; no per-handler change needed (existing `profiler.NewTrace` calls become children via `context.Context` propagation).

## Devops surfaces

- **OTel trace shape change**: exporters now see a top-level `http.server.request` span for every request. Dashboards that filter on the absence of one (unlikely; they look for child spans today) would change.
- **Release note line**: `Added http.server.request OpenTelemetry span at HTTP handler entry. Every inbound request now opens a top-level span with http.request.method, http.route, http.response.status_code, and hecate.request_id attributes; existing per-subsystem spans become children of it through the request context.`

## Test plan

- [x] `go vet ./...` — clean
- [x] `gofmt -s -l .` — clean
- [x] `go test ./internal/api -run TestOTelHTTPSpan -count=1` — 1 / 1 pass (new test, full span shape asserted)
- [x] `go test ./... -race -count=1` — 1780 / 1780 pass across 36 packages

## Why `[skip desktop]`

Pure Go-side middleware addition under `internal/api/`. No `ui/**` or `tauri/**` changes.

### DIFF
--- a/internal/api/middleware.go
+++ b/internal/api/middleware.go
@@ -8,13 +8,19 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"strconv"
 	"time"
 
 	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/propagation"
+	semconv "go.opentelemetry.io/otel/semconv/v1.34.0"
 
 	"github.com/hecate/agent-runtime/internal/telemetry"
 )
+
+var httpServerTracer = otel.Tracer("github.com/hecate/agent-runtime/internal/api")
 
 type middleware func(http.Handler) http.Handler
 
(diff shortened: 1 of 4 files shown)


chicoxyzzy:
Reviewed #151. No blocking findings; `go test ./internal/api -run TestOTelHTTPSpan -count=1` passes, the full `go test ./internal/api` package passes, and `git diff --check` is clean.

I checked the middleware ordering and route/status capture path specifically. The request ID is present before the span starts, `r.Pattern` is read after mux dispatch, and 5xx status propagation goes through the existing recorder/recovery chain.

PR #153, opened 2026-05-17
test(taskstate): consolidate Store conformance behind one entry point [skip desktop]

chicoxyzzy:
## Summary

`internal/taskstate/sqlite_test.go` had 11 top-level `TestSQLiteStore_X` functions with no memory mirror. Nine of them assert Store-interface contracts (round-trips, filter semantics, ordering, cascade-on-delete) that are agnostic to the backend; only two depend on sqlite specifics. Pull the nine into `RunConformanceTests(t, name, factory)` and have both backends invoke it. The memory store previously had zero Store-contract coverage in this package — it now runs the same nine cases as sqlite.

```
3 files changed, 620 insertions(+), 555 deletions(-)
```

## What changes

- **New:** `internal/taskstate/conformance_test.go` (≈470 LOC) — defines `StoreFactory` + `RunConformanceTests`, plus the nine `runStoreX(t, store)` helpers (case bodies lifted verbatim from `sqlite_test.go`).
- **New:** `internal/taskstate/memory_test.go` — one `TestMemoryStoreConformance` calling `RunConformanceTests` with a memory factory.
- **Refactor:** `internal/taskstate/sqlite_test.go` — nine `TestSQLiteStore_X` functions collapse to one `TestSQLiteStoreConformance`. The two sqlite-only tests stay standalone (see below). The `equalStringSlice` / `equalStringMap` helpers stay here because the MCP-round-trip case uses them via the lifted body.

## What moves to conformance (9)

| Test | Asserts |
|---|---|
| `TaskRunStepRoundTrip` | task → run → step round-trip, `step_index ASC` ordering |
| `ListTasksFilterAndLimit` | status filter, limit, `updated_at DESC` ordering |
| `ApprovalRoundTrip` | create → resolve → list approvals on a task |
| `ArtifactRoundTrip` | artifact create + filter by `task_id` / `kind` |
| `RunEventsAppendAndList` | per-run event append, sequence-ASC ordering, `afterSequence` cursor |
| `ListEventsCrossRunFilters` | cross-run lister: event-type set, task-id set, cursor, AND-combine, limit |
| `ListRunsByFilterStatusSet` | run lister with status set + limit |
| `DeleteTaskCascades` | `DeleteTask` removes child runs and steps |
| `TaskMCPServersRoundTrip` | full `MCPServers` slice survives storage round-trip |

## What stays sqlite-only (2)

- **`TestSQLiteStore_RejectsNilClient`** — `NewSQLiteStore(ctx, nil)` constructor validation. No memory analogue (memory's `NewMemoryStore` takes no client).
- **`TestSQLiteStore_BackendName`** — the literal `"sqlite"` string is the assertion, by definition backend-specific. (`Backend() != ""` is implicitly covered by every conformance case running successfully.)

## Memory-vs-sqlite drift

**None surfaced.** All nine lifted cases pass on both backends as-is. Notes for anyone wondering where drift *could* have hidden:

- `ListEvents` defaults `limit` to 200 on both backends and orders sequence-ASC on both — the `limit-caps-the-response-size` subtest exercises an explicit limit, so the default doesn't enter into it.
- The `RFC3339Nano` timestamp gotcha in `sqlite.AppendRunEvent` is sqlite-only mechanics; the round-trip the conformance test asserts (`AppendRunEvent` → `ListRunEvents` → sequence ordering) is uniform.
- `DeleteTask` cascade is enforced in-code on both backends (sqlite via the `DeleteTask` loop over child tables; memory via map deletes scoped to `task_id`). Same observable behavior.

## Out of scope

- The existing `approval_test.go` (3 tests) and `prune_turn_events_test.go` (2 tests) already run against both backends via their own table-driven setup. They're conceptually conformance cases too, but folding them in would interleave unrelated refactor with this consolidation; left as-is.

## Test plan

- [x] `gofmt -s -l .` — clean
- [x] `go vet ./...` — clean
- [x] `go test ./internal/taskstate/... -race -count=1` — 49 / 49 pass (was 32; the increase is the new memory-backend conformance subtests + the existing sqlite conformance now reported as subtests)
- [x] `go test ./... -race -count=1` — 1799 / 1799 pass across 36 packages

## Why `[skip desktop]`

Pure Go-side test refactor under `internal/taskstate/`. No `ui/**` or `tauri/**` changes.


chicoxyzzy:
Reviewed #153. No blocking findings; `git diff --check` is clean, `go test ./internal/taskstate/...` passes, the conformance subtest run passes, and `go test ./...` passes across 36 packages.

I checked the new `RunConformanceTests` harness shape specifically: each top-level conformance case gets a fresh store via the factory, nested event-filter subtests stay sequential on their shared fixture store, and the sqlite-only tests remain isolated in `sqlite_test.go`.

PR #155, opened 2026-05-17
refactor(app): extract settings slice; useRuntimeConsole has zero inline state

chicoxyzzy:
## Summary

The three remaining inline \`useState\` calls in \`ui/src/app/useRuntimeConsole.ts\` (\`settingsConfig\`, \`settingsError\`, \`notice\`) move into a new \`settings\` slice at \`ui/src/app/state/settings.tsx\`, joining the existing 6 slices under \`app/state/\`. After this PR, useRuntimeConsole is pure composition — every piece of state lives in a slice; the hook reads/writes via slice actions only.

## What changes

**New** \`ui/src/app/state/settings.tsx\`:
- \`SettingsState\` = \`{ config, error, notice }\`
- \`SettingsActions\` = \`{ setConfig, updateConfig, setError, setNotice, dismissNotice, dismissNoticeIfMatching }\`
- \`SettingsProvider\` + \`useSettings\` hook following the same pattern as the existing slices

**\`ui/src/app/App.tsx\`** — wraps \`<RuntimeConsoleContextProvider>\` with \`<SettingsProvider>\`.

**\`ui/src/app/useRuntimeConsole.ts\`** — drops the 3 inline \`useState\` declarations, the local \`NoticeState\` type definition, and the inline arrow \`dismissNotice\`. Replaces them with a \`const settings = useSettings()\` integration plus the slice's exported \`NoticeState\` type. The auto-dismiss \`useEffect\` now calls \`settings.actions.dismissNoticeIfMatching(notice)\` rather than passing a functional updater to a React setter.

**\`ui/src/app/useRuntimeConsole.test.tsx\`** — \`SliceProviders\` test wrapper gains \`<SettingsProvider>\` so \`renderHook(() => useRuntimeConsole())\` can find the context.

## API additions on the slice

- \`setConfig(value)\` — replace the config snapshot
- \`updateConfig(updater)\` — functional updater form (used by the optimistic-update + rollback paths in \`deleteProvider\`)
- \`setNotice(value)\` — set or clear the banner
- \`dismissNotice()\` — clear unconditionally (used by the bag's \`actions.dismissNotice\`)
- \`dismissNoticeIfMatching(notice)\` — clear only if the slice still holds the captured notice. Used by the 3s auto-dismiss \`useEffect\` so a newer notice that displaced the timed-out one doesn't get wiped by a stale timeout.

## What doesn't change

- \`useRuntimeConsole\`'s external \`{ state, actions }\` surface is byte-identical: \`state.settingsConfig\`, \`state.settingsError\`, \`state.notice\`, \`actions.dismissNotice\` all stay.
- No view-facing prop or context-shape change. Views that read via \`useRuntimeConsoleContext()\` see exactly the same shape.

## Note on slice scope

The \`notice\` field is a cross-cutting concern (transient banner set by many call sites, not just settings) but lives in the settings slice because shipping a one-field "notice" slice felt heavier than the seam was worth. Documented in the slice's doc comment; if it grows a second non-settings concern, split.

## Test plan

- [x] \`bun run typecheck\` — clean
- [x] \`bunx vitest run\` — 939 / 939 pass across 40 test files
- \`grep "useState\\ |useReducer" ui/src/app/useRuntimeConsole.ts\` — returns 0 hits

## Why no \`[skip desktop]\`?

Touches \`ui/**\`; the desktop bundle re-builds.

### DIFF
--- a/ui/src/app/App.tsx
+++ b/ui/src/app/App.tsx
@@ -7,6 +7,7 @@ import { ChatProvider } from "./state/chat";
 import { ProvidersAndModelsProvider } from "./state/providersAndModels";
 import { RetentionProvider } from "./state/retention";
 import { RuntimeProvider } from "./state/runtime";
+import { SettingsProvider } from "./state/settings";
 import { UsageProvider } from "./state/usage";
 import { usePersistedState } from "../lib/persistedState";
 import { isTauriRuntime } from "../lib/tauri";
@@ -31,9 +32,11 @@ export default function App() {
           <ChatProvider>
             <RetentionProvider>
               <ApprovalsProvider>
-                <RuntimeConsoleContextProvider>
-                  <AppConsole />
-                </RuntimeConsoleContextProvider>
+                <SettingsProvider>
+                  <RuntimeConsoleContextProvider>
+                    <AppConsole />
+                  </RuntimeConsoleContextProvider>
+                </SettingsProvider>
               </ApprovalsProvider>
             </RetentionProvider>
           </ChatProvider>
--- a/ui/src/app/state/settings.tsx
+++ b/ui/src/app/state/settings.tsx
@@ -1,0 +1,113 @@
+// Settings slice: the server-fetched configuration snapshot the
+// settings page renders against, the settings-mutation error string
+// shown next to settings forms, and the transient cross-page notice
+// banner (success / error toast).
+//
+// `notice` is a global UI concern rather than a settings concern —
+// it's set after settings mutations but also after chat actions and
+// other operator-driven events. Lives here because it had no better
+// home and shipping a one-field "notice" slice felt heavier than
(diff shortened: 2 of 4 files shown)


chicoxyzzy:
Reviewed #155. No blocking findings; `git diff --check` is clean, `cd ui && bun run typecheck` passes, `cd ui && bun run test` passes (40 files / 939 tests), and `useRuntimeConsole.ts` has no inline `useState`/`useReducer` declarations left.

I checked the slice migration specifically: App and hook test wrappers include `SettingsProvider`, the legacy `state.settingsConfig/settingsError/notice` and `actions.dismissNotice` surface stays intact through aliases, and the auto-dismiss path preserves the old object-identity guard via `dismissNoticeIfMatching`.

PR #159, opened 2026-05-17
refactor(app): retire useRuntimeConsole facade; views consume slices + coordinators directly

chicoxyzzy:
## Summary

- Retire the `useRuntimeConsole` `{state, actions}` facade from production. Views (13 view files plus AppShell) now consume slice state through `useRuntime` / `useChat` / `useProvidersAndModels` / `useSettings` / `useRetention` / `useApprovals` / `useUsage` and dispatch mutations through coordinator hooks (`useChatActions`, `useWiredProviderActions`, `useAgentAdapterActions`, `usePolicyActions`, `useWiredDashboardActions`, `useRetentionActions`, `useWiredSettingsActions`) directly.
- Delete `ui/src/app/useRuntimeConsole.ts` and `ui/src/app/RuntimeConsoleContext.tsx` from production. The composed hook survives as a test-only composer in `ui/src/test/runtime-console-test-composer.ts`; the 2284-line regression suite moves to `ui/src/test/runtime-console-composition.test.tsx` and exercises that composer end-to-end against real slices + coordinators.
- `App.tsx` mounts the seven slice providers plus a new `<RootEffects />` component (`ui/src/app/state/rootEffects.tsx`) that owns the cross-slice effect cluster the facade used to carry: dashboard load on mount, approvals catch-up on session change, RTK enabled sync, notice auto-dismiss, provider/model default cascade, queued-message drain.

### Derived state hooks

The facade's pre-computed fields move into `ui/src/app/state/derived.tsx`:

- `useChatTarget()` returns the route the active session resolves to.
- `useRuntimeDerivedState()` returns `healthyProviders`, `localProviders`, `cloudProviders`, `localModels`, `cloudModels`, `healthyLocalProviders`, `healthyCloudProviders`, `visibleModels`, `providerScopedModels`, `localProviderIssues`, and `session`.
- `useNewChatAgentID()` returns the agent id surfaced in the sidebar's new-chat button.

A view that needs one calls the per-derivation hook; each derivation runs against slice state, never against fixture state, so there's no parallel source of truth in production.

### Wired coordinator hooks

The cross-slice coordinator hooks accept a per-call params bag the facade used to assemble (`settingsConfig`, `setSettingsConfig`, `loadDashboard`, `setNoticeMessage`, …). To avoid duplicating that assembly in every view, `ui/src/app/state/coordinators/wired.tsx` exposes wired wrappers (`useWiredSettingsActions`, `useWiredDashboardActions`, `useWiredProviderActions`, `useWiredPolicyActions`) that resolve the graph once and return the coordinator bag with no params. Views call them where they would have called the action through the facade. The base `useChatActions` stays parameterised (`{chatTarget, setNoticeMessage}`) because both inputs are derived; views pass them inline.

### Cross-slice forward-dependency

The chat → settings → dashboard cycle (settings `runSettingsMutation` calls `loadDashboard`, which lives on the dashboard coordinator, which composes the chat coordinator's helpers, which need `setNoticeMessage` from the settings coordinator) is resolved in `RootEffects` and in `useWiredDashboardActions` / `useWiredProviderActions` via the same lazy-ref pattern the facade used: the settings coordinator gets a thunk it resolves through a ref, which the dashboard coordinator populates after it's constructed.

## Test plumbing

Tests still want `withRuntimeConsole(ui, {state, actions})` ergonomics and `expect(setProviderAPIKey).toHaveBeenCalledWith(...)`-style assertions. The new wrapper in `ui/src/test/runtime-console-render.tsx`:

- Mounts the seven slice providers; each slice gains an optional `initialState` prop that the providers read once at mount. The seven slice files lose nothing else.
- Builds a `CoordinatorOverrides` value from the fixture's flat `actions` bag, mounted via a new `CoordinatorOverridesProvider`. The slice and coordinator hooks all check this context at the top of their return path through a tiny `applyOverride(real, override)` helper; the override merges per-key over the real implementation so partial overrides fall through cleanly. Production never mounts the provider.
- Renders a `FixtureSyncer` child that re-applies the fixture state on every render. RTL's `rerender(...)` reuses the slice providers (their reducers initialise once at mount), so without the syncer a test that passes new fixture state via `rerender(withRuntimeConsole(ui, { state: nextState, actions }))` would never see slice state update. The syncer mounts outside the overrides context so its dispatches use the real slice actions, not the test's `vi.fn()` stubs.
- Adds two derived-value escapes (`derivedChatTarget`, `derivedNewChatAgentID`) for the previously-precomputed fields tests pin directly. Each hook checks the override first and short-circuits if present.

The test fixture (`runtime-console-fixture.ts`) keeps the same `createRuntimeConsoleFixture` / `createRuntimeConsoleActions` shape — it's now a denormalized snapshot the wrapper fans out into the slice initial states.

## File-by-file

Deleted: `ui/src/app/RuntimeConsoleContext.tsx`, `ui/src/app/useRuntimeConsole.ts`, `ui/src/app/useRuntimeConsole.test.tsx` (renamed to `runtime-console-composition.test.tsx`).

Added: `ui/src/app/state/derived.tsx`, `ui/src/app/state/rootEffects.tsx`, `ui/src/app/state/coordinators/overrides.tsx`, `ui/src/app/state/coordinators/wired.tsx`, `ui/src/test/runtime-console-test-composer.ts` (renamed from `useRuntimeConsole.ts`), `ui/src/test/runtime-console-composition.test.tsx`.

Modified: all seven slice files (add `initialState` prop + override hook check); all seven coordinator files (override merge at return); thirteen view files + `AppShell.tsx` (consume slice + coordinator hooks directly); `App.tsx` (mount `<RootEffects />` instead of `<RuntimeConsoleContextProvider>`); `runtime-console-fixture.ts` (typed against slice types instead of viewmodel); `runtime-console-render.tsx` (slice-provider wrapper + overrides + syncer); `TraceDetail.tsx` (UsageEntry type no longer reads through viewmodel).

## Test plan

- [x] `bun run typecheck` — clean.
- [x] `bunx vitest run` — 939 / 939 pass, same count as master.


chicoxyzzy:
Reviewed #159. No blocking findings.

Verification I ran:
- `git diff --check 449e43f41a374f21362083ea6bc0540ef50b8dc7..HEAD` — clean
- `cd ui && bun run test` — 40 files / 939 tests passed

I checked the facade retirement paths specifically: production imports of `RuntimeConsoleContext` / app-level `useRuntimeConsole` are gone, the remaining `useRuntimeConsole` is confined to the test-only composer, `<RootEffects />` carries the old cross-slice effect cluster, and the coordinator override provider is only used by test scaffolding. I also traced the wired dashboard/provider/settings ref plumbing and did not find a blocking behavior regression.

PR #163, opened 2026-05-17
perf(app): defer getModels to Wave 2 and Usage fetches to view mount

chicoxyzzy:
## Summary

Two surgical changes to the dashboard loader (`ui/src/app/runtimeConsoleDashboard.ts`) that shrink the boot-time fetch footprint and clear the Connecting gate faster.

### Change 1 — drop `getModels()` from Wave 1

Wave 1 (gate-blocking) goes from **4 → 3 parallel calls**: `getHealth`, `getSession`, `getSettingsConfig`. `getModels()` moves to Wave 2 alongside the other parallel calls. `DashboardEssentials` drops its `models` field.

**Trade-off:** the status-bar model count starts at `0 models` briefly (~ 100-300 ms on cold cache) and updates once Wave 2 resolves. Accepted as a deliberate UX choice — the gate clears ~30-50 ms sooner.

### Change 2 — defer Usage fetches to view mount

`getUsageSummary("")` and `getUsageEvents(20)` move **out of Wave 2 entirely**. The dashboard loader no longer fetches them; the slice is no longer hydrated at boot.

A new `useEnsureUsageLoaded()` hook (`ui/src/app/state/usage.tsx`) fetches both on first UsageView mount, dispatches into the slice, and flips a `loaded` flag so in-session navigations don't re-fetch. A full page reload resets the slice and the next UsageView mount re-fetches.

**Savings:** ~80% of boots land on Chats and never visit Usage during the session. For those (the common case), this saves 2 API calls per cold boot — roughly 50-150 ms cumulative depending on backend latency.

## What changes

- `runtimeConsoleDashboard.ts` — Wave 1 = 3 calls; models in Wave 2; usage fetches removed entirely; `DashboardSnapshot` + `DashboardPreviousState` drop `usageSummary` + `usageEvents`.
- `state/coordinators/dashboard.ts` — drops `setUsageSummary` / `setUsageEvents` plumbing + the `useUsage` slice dependency.
- `state/usage.tsx` — adds `loaded: boolean` to state, `markLoaded` action, and the `useEnsureUsageLoaded` hook with inflight-dedup ref + non-throwing failure log via `lib/log.ts`.
- `features/usage/UsageView.tsx` — calls `useEnsureUsageLoaded()` at the top of the component.

## What doesn't change

- `useUsage` slice's `{state, actions}` external shape (other than the new `loaded` field + `markLoaded` action — additive only).
- Behavior when UsageView is mounted: identical end-state.
- Test composer in `ui/src/test/runtime-console-test-composer.ts` — unchanged.
- The 2284-LOC `runtime-console-composition.test.tsx` regression suite — runs unchanged.

## Test plan

- [x] `bun run typecheck` — clean
- [x] `bunx vitest run` — 939 / 939 pass
- One existing assertion in `runtimeConsoleDashboard.test.ts` updated (`essentials.models toEqual []` — `models` is no longer in `essentials`). All other tests pass unchanged.

## Caveat (cosmetic only)

The test runner emits one `usage.ensureLoaded.failed` warning when a test environment doesn't mock the new fetches (the relative URL fails to parse in node-fetch). The warning is non-fatal and tests still pass; can be suppressed in a follow-up by gating the hook on a test env flag if it bothers anyone.

### DIFF
--- a/ui/e2e/usage.spec.ts
+++ b/ui/e2e/usage.spec.ts
@@ -11,3 +11,83 @@ test.beforeEach(async ({ page }) => {
 test("shows the empty usage state when no cloud calls have been recorded", async ({ page }) => {
   await expect(page.locator("text=No cloud usage recorded yet")).toBeVisible();
 });
+
+// The two specs below pin the lazy-fetch contract: Usage data is NOT
+// pulled at boot; UsageView fetches both summary + events on mount
+// and caches via a slice-level `loaded` flag so in-session navigation
+// doesn't re-fetch. A full page reload resets the slice.
+test.describe("Usage lazy-fetch", () => {
+  // Override beforeEach — these specs need to control navigation
+  // themselves to observe boot-time vs view-mount fetch behavior.
+  test.beforeEach(async ({ page }) => {
+    // Replace the default empty-list routes with counters. The default
+    // routes registered in mockGatewayAPIs would have already
+    // satisfied any boot-time fetch, so we re-register OURS on top
+    // (Playwright matches routes in reverse registration order, so
+    // most-recent wins) before navigating.
+    let summaryCount = 0;
+    let eventsCount = 0;
+    await page.route("/hecate/v1/usage/summary*", async route => {
+      summaryCount += 1;
+      await route.fulfill({
+        status: 200,
+        contentType: "application/json",
+        body: JSON.stringify({
+          object: "usage_summary",
+          data: {
+            key: "global",
+            scope: "global",
+            backend: "memory",
+            used_micros_usd: 0,
+            used_usd: "$0.000000",
+          },
+        }),
+      });
+    });
+    await page.route("/hecate/v1/usage/events*", async route => {
+      eventsCount += 1;
+      await route.fulfill({
+        status: 200,
+        contentType: "application/json",
+        body: JSON.stringify({ object: "usage_events", data: [] }),
+      });
+    });
+    // Expose counters to the page-level tests via window-style refs.
+    (page as unknown as { usageRouteCounts: { summary: () => number; events: () => number } })
+      .usageRouteCounts = { summary: () => summaryCount, events: () => eventsCount };
+  });
+
+  test("does not fetch /usage/{summary,events} during boot", async ({ page }) => {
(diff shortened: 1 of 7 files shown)


chicoxyzzy:
Reviewed #163 carefully. No blocking findings.

Verification I ran:
- `git diff --check a8200f7d4169daf3db21be60d504f736de2516cd..HEAD` — clean
- `cd ui && bun run test` — 40 files / 939 tests passed

I traced the boot-loader behavior specifically: `runtimeConsoleDashboard.ts` no longer imports or calls the usage endpoints, Wave 1 now commits only health/session/settings, Wave 2 still writes models before the final dashboard snapshot lands, and `state/coordinators/dashboard.ts` no longer hydrates the usage slice. I also checked `useEnsureUsageLoaded()` for retry/in-flight behavior and `UsageView` for the lazy load handoff; the current full-suite run did not reproduce the warning mentioned in the PR body.

Small non-blocking cleanup I noticed: `runtimeConsoleDashboard.test.ts` still mocks usage endpoints and `emptyPrev` still carries `usageSummary`/`usageEvents`, even though the loader no longer consumes them. Not worth blocking, but trimming that would make future regressions around “dashboard must not fetch usage” easier to spot.

chicoxyzzy:
Re-reviewed after the latest push. No blocking findings.

Verification I reran:
- `git diff --check a8200f7d4169daf3db21be60d504f736de2516cd..HEAD` - clean
- `cd ui && bun run typecheck` - clean
- `cd ui && bun run test` - 41 files / 944 tests passed
- `cd ui && bun run test:e2e usage.spec.ts` - 3/3 passed

The stale dashboard test scaffolding from my earlier note is cleaned up now; the dashboard tests no longer carry the old usage/retention mocks, and the lazy usage e2e coverage pins boot vs first UsageView mount behavior.

Small non-blocking wording note: `useEnsureUsageLoaded` still says it "Dedupes parallel callers via an inflight ref"; that ref is per hook instance, so the comment overstates cross-instance dedupe. Current production only mounts this from `UsageView`, and the new test correctly covers same-hook rerenders, so I do not see this as blocking.

PR #164, opened 2026-05-17
perf(app): defer getProviderPresets to AddProviderModal / TasksView mount

chicoxyzzy:
## Summary

Provider presets move out of the boot-time dashboard loader. The catalog now loads on first \`AddProviderModal\` open (gated by the modal's \`open\` prop) and on first \`TasksView\` mount. A new \`providerPresetsLoaded\` flag on the providersAndModels slice dedupes across in-session re-opens.

## Savings

- **Default-Chats boots** save one API call. ChatView always-mounts the modal (\`open=false\` initially), but the gated hook skips the fetch until the operator clicks "Add provider".
- **Non-Chats boots** save the fetch entirely until either modal or TasksView mounts.

## Boot fanout shape change

| Wave | Before | After |
|---|---|---|
| Wave 1 (gate-blocking) | 3 parallel calls | 3 parallel calls (unchanged) |
| Wave 2 (background) | 5 parallel + 1 conditional | **4** parallel + 1 conditional |

\`DashboardSnapshot\` loses its \`providerPresets\` field. The dashboard coordinator drops the \`setProviderPresets\` plumbing.

## UI rewires

- **\`AddProviderModal\`** calls \`useEnsureProviderPresetsLoaded(open)\` so the hook only fires when the modal is actually open. Critical: ChatView always-mounts the modal, so without the \`open\` gate the optimization would no-op on Chats boots.
- **\`TasksView\`** drops its bespoke local-state fetch and reads \`providerPresets\` from the slice via \`useEnsureProviderPresetsLoaded()\`. The old \`useState\` + inline \`getProviderPresets()\` are gone.

## Slice contract

- New state field: \`providerPresetsLoaded: boolean\` (initial \`false\`).
- New action: \`markProviderPresetsLoaded()\` (idempotent).
- New hook: \`useEnsureProviderPresetsLoaded(when: boolean = true)\` — optional \`when\` gate for always-mounted consumers.

## Test plan

**4 unit tests** (\`providersAndModels.test.tsx\`) cover the hook contract: fetch on first mount, dedup via loaded flag, \`when=false\` gate, failure tolerance, retry-on-next-mount.

**3 e2e tests** (\`provider-presets-lazy.spec.ts\` × 2 + \`boot-shape.spec.ts\` × 1) pin the boot shape (no \`/providers/presets\` fetch at boot on default Chats; exact set of endpoints fired during boot) and the modal-open contract (fetched on first open, not on re-open).

- [x] \`bun run typecheck\` — clean
- [x] \`bunx vitest run\` — 949 / 949 pass
- [x] \`bun run test:e2e provider-presets-lazy.spec.ts\` — 2 / 2 pass, no Vite proxy noise
- [x] \`bun run test:e2e boot-shape.spec.ts\` — 1 / 1 pass

## What doesn't change

- \`useProvidersAndModels\` external shape (additive only — \`providerPresetsLoaded\` + \`markProviderPresetsLoaded\`).
- Existing UsageView, AddProviderModal, ProvidersView, ConnectionsPanel behavior is identical post-fetch.
- Chat coordinators (\`setChatTarget\`, \`deleteProvider\`, rootEffects auto-default cascade) tolerate empty presets gracefully — \`defaultModelForProvider\` falls back to per-provider model list when no preset is found. First cold boot may pick a slightly less optimal default until the user opens AddProviderModal; acceptable trade-off documented in the slice's doc comment.

### DIFF
--- a/ui/e2e/boot-shape.spec.ts
+++ b/ui/e2e/boot-shape.spec.ts
@@ -1,0 +1,71 @@
+import { expect, test } from "./fixtures";
+
+// Pins the exact set of endpoints the UI hits during boot on the
+// default (Chats) workspace. A regression here would flag any change
+// that re-adds a deferred fetch to the dashboard loader OR drops a
+// still-needed one.
+//
+// What's expected at boot today:
+//   Wave 1 (gate-blocking): /healthz, /hecate/v1/whoami, /hecate/v1/settings
+//   Wave 2 (background):    /v1/models, /hecate/v1/agent-adapters,
+//                           /hecate/v1/chat/sessions, /hecate/v1/system/stats,
+//                           /hecate/v1/providers/status (conditional on configured providers)
+//
+// What's expected NOT to fire at boot:
+//   /hecate/v1/usage/summary, /hecate/v1/usage/events  (deferred to UsageView)
+//   /hecate/v1/providers/presets                       (deferred to AddProviderModal / TasksView)
+//   /hecate/v1/retention/*                             (deferred to SettingsView)
+//   /hecate/v1/observability/*                         (deferred to ObservabilityView)
+
+test("default Chats boot fires exactly the expected endpoints, no more", async ({ page }) => {
+  const hits: string[] = [];
+  // Capture every request the page issues; the fixtures' page.route
+  // calls fulfill the responses, but we tap with continue() to keep
+  // both behaviors. Use a request listener instead of additional
+  // route handlers so we don't shadow the fixture routes' fulfillment.
+  page.on("request", req => {
+    const url = new URL(req.url());
+    if (url.pathname === "/healthz"
+      || url.pathname === "/v1/models"
+      || url.pathname.startsWith("/hecate/v1/")) {
+      hits.push(url.pathname);
+    }
+  });
+
+  await page.goto("/");
+  await page.waitForSelector(".hecate-activitybar");
+  await page.waitForSelector(".hecate-activitybar [aria-label^='Chats']");
+  // Give Wave 2 + any settle-down effects a moment to complete.
+  await page.waitForTimeout(500);
+
+  const unique = Array.from(new Set(hits));
+
+  // Endpoints that MUST appear during boot.
+  const expectedAtBoot = [
+    "/healthz",
+    "/hecate/v1/whoami",
+    "/hecate/v1/settings",
+    "/v1/models",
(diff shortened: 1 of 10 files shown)


chicoxyzzy:
Reviewed #164 carefully. No blocking findings.

Verification I ran:
- `git diff --check a8200f7d4169daf3db21be60d504f736de2516cd..HEAD` — clean
- `cd ui && bun run test` — 42 files / 949 tests passed
- `cd ui && bun run test:e2e provider-presets-lazy.spec.ts` — 2 / 2 passed
- `cd ui && bun run test:e2e usage.spec.ts` — 3 / 3 passed

I traced the new preset path specifically: `runtimeConsoleDashboard.ts` no longer imports/calls `getProviderPresets`, the dashboard coordinator no longer writes presets into the slice, `AddProviderModal` correctly gates `useEnsureProviderPresetsLoaded(open)`, and `TasksView` reads the shared slice catalog instead of local `getProviderPresets()` state.

Non-blocking notes: the hook comment says it dedupes parallel callers via the in-flight ref, but that ref is per hook instance; the real dedupe across surfaces happens after `providerPresetsLoaded` flips. Current mounted surfaces make that okay, just worth wording precisely. The targeted provider-presets e2e also still prints Vite proxy noise for adapter probes/grants when it navigates to Connections, though the assertions pass.

chicoxyzzy:
Re-reviewed after the latest push. No blocking findings.

Verification I reran:
- `git diff --check a8200f7d4169daf3db21be60d504f736de2516cd..HEAD` - clean
- `cd ui && bun run typecheck` - clean
- `cd ui && bun run test` - 42 files / 949 tests passed
- `cd ui && bun run test:e2e provider-presets-lazy.spec.ts` - 2/2 passed
- `cd ui && bun run test:e2e boot-shape.spec.ts` - 1/1 passed
- `cd ui && bun run test:e2e usage.spec.ts` - 3/3 passed

The in-flight dedupe wording I flagged earlier is addressed in `providersAndModels.tsx`; it now accurately calls out the per-hook ref and loaded-flag behavior. The new boot-shape spec also uses request listeners rather than route shadowing, which is the right shape for pinning the lazy-fetch contract.

Remaining non-blocking cleanup: `provider-presets-lazy.spec.ts` still prints Vite proxy noise for adapter probe/grants requests while navigating Connections, although assertions pass. Also, the PR body still says 948 tests, while this head reports 949.

PR #195, opened 2026-05-22
feat(projects): scope chats by project

chicoxyzzy:
## Summary

- add durable project ownership to chat sessions and expose project-scoped chat filtering in the UI
- add a compact projects rail with create, rename, delete, and project-filtered chat lists
- remove default model fallbacks so Hecate chats start from explicit operator/provider state
- allow empty Hecate chat shells so starting a new Hecate chat does not leave stale external-agent controls visible

## Why

Projects are the next foundation for per-project context and memory. This keeps chats usable both inside a selected project and outside any project, while making the chat shell consistent: creating a Hecate chat should select a Hecate-owned session immediately, even before a model/workspace is chosen.

## Validation

- `GOCACHE="$PWD/.gocache" go test ./internal/api ./internal/chat`
- `GOCACHE="$PWD/.gocache" go vet ./internal/api ./internal/chat`
- `cd ui && bun run test -- ChatView runtime-console-composition`
- `cd ui && bun run test`


chicoxyzzy:
@copilot please review again. The latest push addresses the project-delete cascade, UI failed-delete handling, and docs lifecycle comments.

PR #207, opened 2026-05-24
add structured agent tools

chicoxyzzy:
## Summary

Adds more built-in `agent_loop` tools so agents can inspect and edit workspaces without falling back to broad shell/git commands:

- Adds structured workspace tools: `grep`, `glob`, ranged `read_file`, and `apply_patch`.
- Adds artifact and git inspection tools: `artifact_read`, `git_status`, and `git_diff`.
- Updates approval-policy mapping so structured tools are covered by the existing `read_file`, `file_write`, `git_exec`, and `all_tools` gates.
- Documents the expanded built-in tool catalog and patch/review behavior.

## Validation

- `go test ./internal/orchestrator`
- `go vet ./...`
- `GOCACHE="$(pwd)/.gocache" go test -race -timeout 10m ./...`
- `git diff --check`


chicoxyzzy:
Addressed the review feedback: ranged read_file now scans the full file before applying max_bytes, git_diff disables textconv and pager behavior, and the docs/tool copy now match the current tool count and approval policy. The failed Windows bootstrap check was a setup-go transient; rerun is green now. Copilot, please take another look.

chicoxyzzy:
Addressed the latest review threads: apply_patch now tolerates blank separator lines without weakening patch line prefixes, git read tools drain command output through a capped buffer instead of CombinedOutput, and inline patch events emit bytes_written consistently as an int. Added focused tests for each behavior and ran go test ./internal/orchestrator, go vet ./internal/orchestrator, and git diff --check. Copilot, please take another look.

chicoxyzzy:
Addressed the latest review threads: apply_patch now denies non-proposed writes in read-only runs, validates patch operation paths before resolving them, and caps update/delete target size before reading. Added focused tests and ran `go test ./internal/orchestrator`, `go vet ./internal/orchestrator`, and `git diff --check`. Copilot, please take another look.

chicoxyzzy:
Addressed the latest git environment review: git_status/git_diff now run git with the same sanitized environment used by sandbox-spawned tools, and there is a regression test with a fake git binary proving non-allowlisted secrets are not inherited. Ran `go test ./internal/orchestrator`, `go vet ./internal/orchestrator`, and `git diff --check`. Copilot, please take another look.

chicoxyzzy:
Addressed the latest review threads.

- `apply_patch` now resolves each operation through the symlink-aware workspace join before preparing or writing, so existing symlink components are rejected preflight.
- Ranged `read_file` now rejects files over the hard cap before scanning and uses a bounded line reader instead of unbounded `ReadString` allocation.
- Added targeted tests for symlink escape prevention and oversized ranged reads.

Verification run locally:
- `go test ./internal/orchestrator`
- `go vet ./internal/orchestrator`
- `git diff --check`

Copilot, please take another look.

PR #211, opened 2026-05-25
feat(runtime): add workspace-bound process and git runners

chicoxyzzy:
## Summary

- Add WorkspaceFS as the shared workspace-bound filesystem layer for Hecate-controlled reads, writes, patch targets, and ACP terminal file callbacks.
- Add ProcessRunner and GitRunner seams so sandbox execution and internal git operations go through bounded, testable runners instead of ad hoc process calls.
- Route agent loop tools, sandbox writes, chat/task patch helpers, changed-file revert, and adapter diff capture through the new runtime seams.
- Document the WorkspaceFS/GitRunner security posture in `docs/security.md`.

## Validation

- `rtk go test -race -timeout 10m ./...`
- `rtk go vet ./...`
- `rtk just docs-format-check`

## Notes

This is intentionally runtime/security infrastructure work. It does not change the operator-facing chat flow, but it makes file/process/git boundaries explicit and easier to audit.


chicoxyzzy:
@copilot please review this PR again.

PR #214, opened 2026-05-25
Guard task patch revert conflicts

chicoxyzzy:
## Summary

Adds a safety precondition before reverting task patch artifacts. Revert now verifies that the current workspace file still matches the patch artifact's expected after-content before writing the original content back, so operator or agent edits made after patch application are not silently clobbered.

This also handles create/delete patches explicitly:

- Modified-file patches only revert when the current file content matches the expected after-content.
- New-file patches still revert by removing the file.
- Deleted-file patches now correctly treat a missing current file as the expected applied state and restore the original before-content.
- Drifted, missing, already-reverted, or otherwise changed targets return `409 Conflict` and leave the workspace unchanged.

## Tests

- `go test ./internal/api -run 'TestTaskRunPatchRevert|TestVerifyPatch(Revert|Apply)Precondition|TestPatchContentExtractsBeforeAndAfter'`
- `go vet ./internal/api`


chicoxyzzy:
## Review

The broader of the two patch-revert PRs (this one and #218). **They overlap directly on `handler_tasks_patches.go` and `server_test.go` and can't both merge.** This one is the more complete implementation — recommend keeping it and closing #218.

### Strengths

- Refactoring `patchAfterContent` to return `(content, existed, err)` and computing `existed` from the diff header (`+++ /dev/null`) is the right move — the patch artifact stays the source of truth, not the filesystem state.
- `verifyPatchRevertPrecondition(fsys, rel, after, afterExisted)` correctly picks the precondition per patch kind: existing file → content must match; deleted-file patch → target must be absent. The `!afterExisted && errors.Is(err, os.ErrNotExist)` → nil branch is tested.
- Reusing `patchContentMatches` in both `verifyPatchApplyPrecondition` and `verifyPatchRevertPrecondition` avoids a future drift between the two paths on the unified-diff final-newline edge case.
- Test coverage is solid across all three patch kinds (modified / new-file / deleted-file), plus repeat-revert idempotency and a focused unit test for the precondition helper.

### Concerns

- **No documentation update.** Unlike #218, this PR doesn't touch `docs/events.md` or `docs/runtime-api.md`. The new `409 Conflict` behavior and the "no `tool.file.reverted` event on conflict" guarantee are operator-visible and should be written down. Easiest path: lift the doc paragraphs straight from #218 before merging.
- `patchContentMatches` only normalizes "expected ends with `\n`, current doesn't". The reverse (current has `\n`, expected doesn't) is rejected — which is the right call, but a table-driven test like #218's `TestPatchContentMatchesAllowsOnlyFinalNewlineAmbiguity` would lock that intent in. Minor.
- `verifyPatchRevertPrecondition` and `verifyPatchApplyPrecondition` now share the same error-categorization shape; could be a small helper, but fine as-is.

### Verdict

Mergeable. Suggested order: port #218's doc paragraphs into this PR, then close #218.

chicoxyzzy:
Addressed the review follow-up here: added the revert-conflict docs to docs/events.md and docs/runtime-api.md, and added the final-newline ambiguity test so the one accepted asymmetric case is explicit.\n\nVerification run:\n- go test ./internal/api\n- go vet ./internal/api\n- git ls-files -z '*.md' '*.mdc' | xargs -0 ui/node_modules/.bin/oxfmt --check

PR #215, opened 2026-05-25
Refresh provider discovery cache on demand

chicoxyzzy:
## Summary
- Add explicit provider/model refresh paths that bypass completed discovery cache entries while preserving normal cached reads.
- Share same-provider in-flight discovery work so concurrent cache misses do not duplicate upstream /v1/models requests.
- Document the refresh=true query flag for provider status and model listing endpoints.

## Validation
- go build ./...
- go test ./internal/providers
- go test ./internal/api -run 'TestProviderStatusAndModelsRefreshQueryBypassesProviderCache'
- go test ./internal/providers ./internal/catalog ./internal/gateway ./internal/api
- go vet ./internal/providers ./internal/catalog ./internal/gateway ./internal/api
- GOCACHE="/Users/chicoxyzzy/dev/hecate-provider-discovery-refresh-singleflight/.gocache" go test -race -timeout 10m ./...

### DIFF
--- a/docs/runtime-api.md
+++ b/docs/runtime-api.md
@@ -519,6 +519,11 @@ Runtime provider readiness snapshot. The UI uses this endpoint to explain
 whether a configured provider can receive traffic right now and why it may be
 skipped by routing.
 
+Pass `refresh=true` or `refresh=1` when the operator explicitly asks to
+refresh provider discovery. Normal reads keep using the provider capability
+cache; explicit refresh bypasses the completed cache while still sharing any
+same-provider discovery request already in flight.
+
 ```json
 GET /hecate/v1/providers/status
 → 200
@@ -655,6 +660,10 @@ operator adds a provider.
 Lists models currently known to configured providers. Each row includes Hecate
 metadata under `metadata`, including the effective model capability snapshot
 used by the Chats target picker.
+
+Pass `refresh=true` or `refresh=1` for an explicit operator refresh. Without
+that query parameter, the endpoint keeps normal provider discovery cache
+behavior.
 
 ```json
 GET /v1/models
--- a/internal/api/handler.go
+++ b/internal/api/handler.go
@@ -550,6 +550,7 @@ func (h *Handler) HandleModels(w http.ResponseWriter, r *http.Request) {
 }
 
 func requestRefresh(r *http.Request) bool {
+	// Accept explicit booleans only so accidental ?refresh does not bypass cache.
 	raw := strings.TrimSpace(r.URL.Query().Get("refresh"))
 	return raw == "1" || strings.EqualFold(raw, "true")
 }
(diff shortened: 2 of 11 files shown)


chicoxyzzy:
## Review

Clean change. Optional `CapabilityRefresher` interface with defensive fallbacks at each layer, `?refresh=true|1` query plumbed through Handler → Service → Catalog → Provider, plus singleflight-style in-flight dedup for capability discovery. Docs updated.

### Strengths

- **Defensive layering.** A stub provider that doesn't implement `RefreshCapabilities` quietly falls back to `Capabilities`, and the `refreshableCatalog` type assertion in `service_providers.go` does the same for the catalog. New code can't break older test stubs.
- **Sync correctness in `resolveCapabilities`.** `call.caps` is written under the mutex before `close(call.done)`; waiters read `call.caps` after `<-call.done`, so the channel close serves as the happens-before edge. Both the success and error-fallback paths clear `*inFlight = nil` and close `done` before unlocking — no goroutine leak.
- **Tests are thorough.** Cached read, manual refresh bypass, error→cached fallback, expiry→rediscover, 8-goroutine concurrent miss sharing one flight, 8-goroutine concurrent refresh sharing one flight.

### Concerns

- **Mixed normal+refresh semantics.** If caller A starts a normal discovery and is mid-flight, a concurrent caller B with `refresh=true` hits the `*inFlight != nil` branch and piggybacks on A's result. So B gets fresh-ish data from a flight started milliseconds earlier rather than a brand-new upstream call. Defensible (B gets the latest discovered state, just from a flight that started a moment ago) but not what "bypass the cache" literally promises. Worth either a comment in `capabilities_cache.go` calling this out or a `TestResolveCapabilitiesRefreshPiggybacksOnInFlightNormalCall` to lock in the intent.
- **`requestRefresh` is strict.** Only accepts `?refresh=1` or `?refresh=true` (case-insensitive). Empty-value `?refresh`, `?refresh=yes`, `?refresh=on` are silently treated as no-refresh. Docs match the implementation, so internally consistent — but if any UI/CLI generates these URLs, double-check it uses one of the two accepted forms. A one-line comment in `requestRefresh` naming the accepted forms would help future readers.
- **Error path observability on manual refresh.** When `discover` fails and the fallback `staticCapabilities("config")` is cached for `retryAfter`, a manual-refresh waiter receives the fallback caps with no error. The fallback carries `LastError`, so a careful caller can detect it — UI should be wired to surface that on the refresh button so an operator's explicit refresh doesn't silently look successful while upstream actually failed.
- **Catalog interface check via runtime assertion.** `s.catalog.(refreshableCatalog)` works, but there's only one `Catalog` implementation; consider hoisting `SnapshotRefresh` into the `Catalog` interface itself in a follow-up. Refactor candidate, not a blocker.

### Verdict

Ready to merge after (optionally) the mixed-flight comment/test and the `requestRefresh` parsing note.

PR #218, opened 2026-05-25
fix(tasks): prevent stale patch reverts

chicoxyzzy:
## Summary
- verify the current workspace file still matches the patch artifact before reverting
- return HTTP 409 and leave files unchanged if an operator or agent edited the target after patch application
- keep repeated/already-reverted behavior intact and cover changed-target rejection with an API test

## Validation
- `go test ./internal/api`
- `go vet ./internal/api`

### DIFF
--- a/docs/events.md
+++ b/docs/events.md
@@ -447,7 +447,7 @@ Emitted when an operator calls the patch apply endpoint for a proposed patch art
 
 ### `tool.file.reverted`
 
-Emitted when an operator calls the patch revert endpoint. The file is restored from Hecate's own patch artifact and the artifact status changes from `applied` to `reverted`.
+Emitted when an operator calls the patch revert endpoint. The file is restored from Hecate's own patch artifact and the artifact status changes from `applied` to `reverted`. The revert endpoint first verifies that the current file still matches the patch artifact's captured after-content; if the file drifted, the request returns `409 conflict` and no revert event is emitted.
 
 If the current workspace file no longer matches the patch artifact's expected after-content (or the file exists when reverting a patch that created a new file), the revert endpoint returns `409 Conflict`, leaves the workspace unchanged, and does not emit `tool.file.reverted`.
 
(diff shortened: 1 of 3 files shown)


chicoxyzzy:
## Review

This and #214 fix the same bug (stale-target patch revert) and conflict directly on `handler_tasks_patches.go` and `server_test.go`. **Recommend closing this PR in favor of #214** and porting the docs over.

### Why #214 wins

- This PR's `verifyPatchRevertPrecondition` doesn't take an `afterExisted` flag, so any `ReadFile` error is unconditionally treated as drift. That means reverting a **delete-file patch** (where the artifact's `+++ /dev/null` header says "file should now be absent") will always return 409 — even when the file is correctly missing. The `TestTaskPatchRevertRejectsMissingTarget` test codifies that behavior, so it's intentional, but it's a partial fix.
- #214 handles all three patch kinds (modified / new-file / deleted-file), refactors `patchAfterContent` to return the `existed` flag from the diff header, and shares `patchContentMatches` between apply and revert so the two paths can't drift apart on the final-newline edge case.

### What this PR does better than #214

- **Docs.** The `docs/events.md` and `docs/runtime-api.md` paragraphs explaining the 409 behavior and the "no `tool.file.reverted` event on conflict" guarantee are good and should be carried over to #214.
- **Table-driven `patchContentMatches` test.** `TestPatchContentMatchesAllowsOnlyFinalNewlineAmbiguity` is worth lifting into #214 too — it locks the helper's intent in a way #214's coverage doesn't.

### Nit

- `fmt.Errorf("patch target changed before revert: %w", err)` will leak `os.PathError` strings (with absolute workspace paths) into the 409 response body. `verifyPatchApplyPrecondition` does the same, so it matches existing style — but worth a future cleanup.

### Verdict

Close this PR; port the docs and the `patchContentMatches` table test into #214.

chicoxyzzy:
The useful docs/test pieces from this PR have been ported into #214, which now contains the fuller revert-conflict implementation and coverage. Treating this PR as superseded unless you want it kept open for comparison.

chicoxyzzy:
Resolved the conflicts by rebasing onto origin/master. The original conflicting revert implementation was superseded by the fuller patch-revert safety work already on master, so I kept master’s implementation and removed the duplicate matcher test from this branch.\n\nVerification run:\n- go test ./internal/api\n- go vet ./internal/api\n- git ls-files -z '*.md' '*.mdc' | xargs -0 ui/node_modules/.bin/oxfmt --check

PR #224, opened 2026-05-26
fix(ui): polish external agent readiness

chicoxyzzy:
## What changed
- Shares one external-agent readiness helper between Connections and Chat so discovery, probe, auth, billing, and setup states are interpreted consistently.
- Adds adapter-specific repair copy for Cursor Agent and Grok Build, including the exact local sign-in commands and the Grok Build model-selection requirement.
- Quiets transcript runtime metadata by keeping run/duration visible while moving native session details into hover text.
- Adds restart reconciliation coverage for an external-agent session that is restored as running without a live assistant message.
- Broadens fake browser coverage for Cursor Agent and Grok Build setup states, without launching real external tools.

## Commits
- `fix(ui): share external agent readiness`
- `fix(ui): clarify adapter setup repairs`
- `fix(ui): quiet transcript runtime metadata`
- `test(chat): cover external session restart reconciliation`
- `test(ui): broaden external agent readiness e2e`

## Tests run
- `go test ./internal/chat`
- `go vet ./internal/chat`
- `cd ui && bun run test`
- `cd ui && bunx playwright test ui/e2e/chat.spec.ts --grep "external-agent chat uses|external-agent running|Cursor Agent setup|Grok Build setup|Claude Code setup stays"`

## Non-goals
- No new settings panels.
- No real external CLIs are invoked by tests.
- No broad chat architecture rewrite.

### DIFF
--- a/internal/chat/store_test.go
+++ b/internal/chat/store_test.go
@@ -407,4 +407,31 @@ func runStoreReconcileInterruptedRuns(t *testing.T, store Store) {
 	if count != 0 {
 		t.Fatalf("second reconciled count = %d, want 0", count)
 	}
+
+	orphaned, err := store.Create(ctx, Session{
+		ID:              "chat_orphaned_external",
+		Title:           "Orphaned external run",
+		AgentID:         "grok_build",
+		DriverKind:      "acp",
+		NativeSessionID: "native_orphaned",
+		Workspace:       "/tmp/hecate",
+		Status:          "running",
+	})
+	if err != nil {
+		t.Fatalf("Create(orphaned): %v", err)
+	}
+	count, err = ReconcileInterruptedRuns(ctx, store, now.Add(2*time.Second))
+	if err != nil {
+		t.Fatalf("ReconcileInterruptedRuns orphaned: %v", err)
+	}
+	if count != 1 {
+		t.Fatalf("orphaned reconciled count = %d, want 1", count)
+	}
+	got, ok, err = store.Get(ctx, orphaned.ID)
+	if err != nil || !ok {
+		t.Fatalf("Get(orphaned): ok=%v err=%v", ok, err)
+	}
+	if got.Status != "cancelled" {
+		t.Fatalf("orphaned session status = %q, want cancelled", got.Status)
+	}
 }
--- a/ui/e2e/chat.spec.ts
+++ b/ui/e2e/chat.spec.ts
@@ -2939,6 +2939,147 @@ type ClaudeAdapterFixture = {
   healthStatus?: "ready" | "auth_required" | "not_installed" | "error";
 };
 
+type ExternalAdapterFixture = ClaudeAdapterFixture & {
+  id: string;
+  name: string;
+  command: string;
+  args?: string[];
+};
+
+async function openExternalAgentReadinessFixture(page: Page, fixture: ExternalAdapterFixture) {
+  const sessionID = `${fixture.id}-onboarding-e2e`;
+  const adapter = {
+    id: fixture.id,
+    name: fixture.name,
+    kind: "acp",
+    command: fixture.command,
+    args: fixture.args,
+    available: fixture.available ?? true,
+    status: fixture.available === false ? "missing" : "available",
+    error: fixture.available === false ? `${fixture.command} command not found` : undefined,
+    description: `Run ${fixture.name} through ACP as a long-lived external coding-agent session supervised by Hecate.`,
+    cost_mode: "external",
+    auth_status: fixture.authStatus ?? "unknown",
+    auth_error:
+      fixture.authStatus === "unauthenticated"
+        ? `Run ${fixture.command.replace(/\s+.*/, "")} login`
+        : undefined,
+  };
+  const status = fixture.available === false ? "not_installed" : (fixture.healthStatus ?? "ready");
+
+  await page.route("/hecate/v1/agent-adapters*", async (route) => {
+    const method = route.request().method();
+    const url = route.request().url();
+    if (method === "POST" && url.includes(`/${fixture.id}/probe`)) {
+      await route.fulfill({
+        status: 200,
+        contentType: "application/json",
+        body: JSON.stringify({
+          object: "agent_adapter_probe",
+          data: {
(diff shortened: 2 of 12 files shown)


chicoxyzzy:
## Review

All required CI green (Tauri jobs skipped, as expected for a UI-only PR).

### Overview

Five small, related changes bundled together:

1. **Extract `ui/src/lib/external-agent-readiness.ts`** — pulls the discovery+probe+auth+billing+setup state machine out of `ConnectionsPanel.tsx` and `ChatView.tsx` into one `resolveExternalAgentReadiness()` helper returning `{ kind, tone, label, needsRepair, loginCommand, setupHint, detail, authStatus, authError, verifiedByProbe }`. ConnectionsPanel loses 143 lines of local logic; `ChatView` drops its bespoke `needsExternalAgentSetup` check.
2. **Adapter-specific copy** for Cursor Agent and Grok Build — per-ID `externalAgentLoginCommand`, `externalAgentSetupHint`, and `externalAgentSetupMessage`. Grok Build's copy correctly mentions the model-selection requirement.
3. **Quieter transcript runtime metadata** — splits `formatAgentRuntimeMeta` so the visible line is just `Run … · 2.0s`; the native ACP session id moves into the `title=` hover attribute via a new `runtimeMetaTitle` prop on `TranscriptMessageRow`.
4. **New Go reconcile test** for an external-agent session restored as `running` without a live assistant message — asserts `ReconcileInterruptedRuns` returns `count == 1` and flips the session to `cancelled`.
5. **New Playwright fixture** `openExternalAgentReadinessFixture` covering Cursor Agent and Grok Build setup states with all HTTP intercepted; no real CLIs spawned.

### Strengths

- **The extracted helper is the right shape.** Pure, no React, no side effects, well-typed (effective discriminated union via `kind`), and has its own unit tests covering the three high-value branches (probe overrides stale auth, auth probe → sign-in, missing CLI → setup with adapter-specific copy).
- **Per-adapter copy is centralized.** Adding the next adapter is one `case` in each of `externalAgentLoginCommand` / `externalAgentSetupHint` / `externalAgentSetupMessage`. No more drift between Connections and Chat copy.
- **Consumer call sites are cleaner.** `ChatView` becomes `resolveExternalAgentReadiness(...).needsRepair` — one boolean, one source of truth. `ConnectionsPanel`'s `adapterStatusDetail` shrinks to ~10 lines.
- **Transcript change is a real UX win.** Run id + duration are what an operator needs on every line; the native ACP session id is diagnostic and rarely needed. Moving it to `title=` keeps it one hover away. The test updates assert both behaviors explicitly.
- **Reconcile test catches a real post-restart edge case** and asserts both the count and the final session status. Lives in the per-store helper so it runs against memory + sqlite.
- **E2E fixture honors the "no real external tools" non-goal.** All adapter discovery, probe, and chat-session traffic is intercepted with `page.route`; the tests verify operator-facing copy, not adapter behavior.

### Concerns

**1. Unprobed-but-available adapters now claim "ready".** The old `adapterStatusState` returned `null` when no condition matched (e.g. `available=true`, no health record yet, `auth_status="unknown"`), and `ConnectionsPanel` rendered no chip in that case via `{state && (…)}`. The new helper always returns a value and falls through to:

```ts
return { kind: "ready", tone: "green", label: "ready", needsRepair: false, … };
```

So an adapter that's available, has unknown auth, and hasn't been probed will now render a green "ready" chip. The ConnectionsPanel JSX was updated accordingly to drop the `{state && (` guard.

This is defensible ("installed, nothing visibly broken, let the operator try"), but it's a real behavior change for the "haven't probed yet" first paint, and the unit tests don't lock it in. Worth either a one-line comment in `resolveExternalAgentReadiness` explaining the fall-through, or a test case `it("treats available adapter with unknown auth and no probe as ready", …)`. If unintended, the fall-through should return a softer state (`kind: "issue"` with `tone: "muted"`, label like "not verified" or similar) until the probe lands.

**2. `detail` for `sign_in` falls through to install copy.** In the `sign_in` branch:

```ts
detail: health?.hint || authError || setupHint
```

For Cursor Agent with `auth_status="unauthenticated"` and no health record, `authError` may be empty and `detail` falls through to `setupHint` (\"Install Cursor's command-line agent, confirm cursor-agent is on PATH, then run cursor-agent login.\"). The `kind` is `sign_in` (we already know the CLI is installed; just sign in), so installation instructions read a little off — though the recovery path does include the same login command at the end, so it's not actively wrong. Tightening this would mean a separate `signInHint` per adapter, or trimming the install preamble when `adapter.available` is true.

**3. `isSetupProbe` relies on health-error string matching.** The `text.includes(\"app cli missing\") || text.includes(\"command was not found\") || text.includes(\"setup docs:\") || text.startsWith(\"install \")` heuristic is inherited from the old code, so this PR didn't introduce it — but moving it into a shared helper increases the blast radius if the gateway's error phrasing ever shifts. Consider a structured `health.kind` or `health.reason` field on the gateway side as a follow-up; for now, a comment naming the gateway source-of-truth function would help.

**4. Native `title=` attribute for transcript hover.** `runtimeMetaTitle` is rendered as a plain `title=` on a `<span>`. Native tooltips aren't keyboard-accessible and screen-reader behavior is inconsistent. For diagnostic info (native ACP session id) this is acceptable, and the test only checks `toHaveAttribute(\"title\", …)`. If an a11y audit later flags this, migrate to a real tooltip primitive. Not a blocker.

**5. Small redundancy in `adapterStatusDetail`.** Still takes `visibleHealthError` even though `readiness.detail` already incorporates `health?.hint`. The function uses `readiness.detail || visibleHealthError` as a fallback when readiness produced no detail. Works, but slightly belt-and-braces. Could be folded into `resolveExternalAgentReadiness` if you want one source of truth for the rendered message; defensible to leave it where it is because Chat doesn't surface `visibleHealthError`.

### Verdict

Solid refactor + targeted UX polish. The helper extraction is well-scoped and the new tests cover the substantive paths. **Recommend merging after confirming concern #1** — either add a comment/test pinning the "unprobed available adapter = ready" decision, or soften that fall-through to something less assertive. Everything else is nice-to-have.

PR #230, opened 2026-05-26
feat(projects): add project context foundation

chicoxyzzy:
## Summary

- Add durable project context-source metadata to the projects API and memory/SQLite stores.
- Include enabled project context-source metadata in chat context packets for direct model, Hecate task-backed, and External Agent turns.
- Update runtime/RFC docs to clarify that this is provenance metadata only; source contents are not injected yet.

## Validation

- `go test ./internal/api ./internal/projects`
- `go vet ./internal/api ./internal/projects`
- `just docs-format-check`
- `git diff --check`

## Notes

This is the foundation for future project-scoped context and memory. It keeps the current behavior intentionally conservative: context packets expose configured project sources for inspection, but Hecate does not read or inject those files yet.

### DIFF
--- a/docs/rfcs/context-assembly-and-injection-boundaries.md
+++ b/docs/rfcs/context-assembly-and-injection-boundaries.md
@@ -101,6 +101,10 @@ Candidate sources are collected from the active runtime surface:
 | External imports          | Future imported Codex/Claude transcripts, PR comments, issue text, web content, raw adapter output. |
 
 Discovery is allowed to over-collect candidates. Later stages decide inclusion.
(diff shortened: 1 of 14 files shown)


chicoxyzzy:
Round 3 is addressed. The cross-backend ordering concern is now closed at the model layer: context sources are normalized to the same stable order memory and SQLite expose (`enabled DESC, path ASC, id ASC`), with mirrored memory/SQLite tests using a fixture that would have diverged before.

Current state: branch is clean, PR merge state is clean, and CI is green. I think this is ready to merge.

PR #232, opened 2026-05-26
fix(agent): use ACP model controls for Grok Build

chicoxyzzy:
## What changed
- Surface Grok Build ACP model state as the chat model control and apply model changes through ACP `session/set_model`.
- Stop treating Grok Build model selection as a launch-required `--model` control.
- Stop exposing Grok Build reasoning as a Hecate-managed launch control, since changing launch-only reasoning would require closing and recreating the native ACP session.
- Generalize the ACP model-state bridge so any built-in ACP adapter that reports models can surface the same model control.

## Why
Grok Build now reports models through ACP. Using that session-native control avoids unnecessary adapter restarts and removes the confusing model-required state before the ACP session is prepared. Reasoning is not surfaced by Hecate unless an adapter reports it as a live ACP session control.

## Tests
- `rtk go test ./internal/agentcontrols ./internal/agentadapters ./internal/api`
- `rtk go vet ./internal/agentcontrols ./internal/agentadapters ./internal/api`
- `rtk bash -lc "git ls-files -z '*.md' '*.mdc' | xargs -0 ./ui/node_modules/.bin/oxfmt --check"`
- `rtk bun scripts/check-mermaid.ts`
- `rtk git diff --check`

## Non-goals
- No UI layout changes.
- No real external CLI calls in tests.

### DIFF
--- a/internal/agentadapters/registry.go
+++ b/internal/agentadapters/registry.go
@@ -295,38 +295,14 @@ func BuiltIns() []Adapter {
 			SupportedRange: ">=0.1.0",
 		},
 		{
-			ID:               "grok_build",
-			Name:             "Grok Build",
-			Command:          "grok",
-			Args:             []string{"agent"},
-			LaunchSuffixArgs: []string{"stdio"},
+			ID:      "grok_build",
+			Name:    "Grok Build",
+			Command: "grok",
+			Args:    []string{"agent", "stdio"},
 			CandidatePaths: []string{
 				"${HOME}/.local/bin/grok",
 				"/opt/homebrew/bin/grok",
 				"/usr/local/bin/grok",
-			},
-			LaunchOptions: []LaunchSelectConfig{
-				// Grok Build exposes model selection through ACP session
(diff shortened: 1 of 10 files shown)


internal/agentadapters/registry.go
```diff
@@ -321,11 +316,12 @@ func BuiltIns() []Adapter {
 					UnsetDescription: "Use Grok Build's default reasoning effort.",
 					ArgTemplate:      []string{"--reasoning-effort", "{reasoning_effort}"},
 					Options: []LaunchSelectOption{
+						{ID: "none", Name: "None"},
+						{ID: "minimal", Name: "Minimal"},
 						{ID: "low", Name: "Low"},
 						{ID: "medium", Name: "Medium"},
 						{ID: "high", Name: "High"},
 						{ID: "xhigh", Name: "XHigh"},
-						{ID: "max", Name: "Max"},
```
chicoxyzzy, line 328: can we load those using ACP as well and maybe re-use that logic for all the ACP external agents?

PR #234, opened 2026-05-26
feat(chat): add workspace tools

chicoxyzzy:
## Summary
- Add workspace open actions for chat workspaces through the desktop command or the local loopback endpoint.
- Add a workspace changes side panel for chat-level Git diffs, with per-file inspection and guarded discard flows.
- Improve transcript activity rendering for changed files, command output, and captured diff previews.
- Persist the shared chat right-panel width and version the workspace opener preference.

## Review follow-up
- Reject `X-Forwarded-For` and `X-Real-IP` on local workspace UI endpoints and document the local-only proxy boundary.
- Add workspace-open launcher error-path coverage for both native and local-gateway dispatch.
- Add the italic-with-inline-code markdown fixture.
- Add tests for right-panel width persistence and workspace opener preference migration.

## Tests
- `cd ui && bun run lint`
- `cd ui && bun run test`
- CI is green, including Go race, Tauri Rust tests, TypeScript E2E, Ollama E2E, and desktop bundle builds.


chicoxyzzy:
## Review

Large PR (~78 files). Reads cleanly overall; the new `workspace_open` and `chat_files` endpoints are the meaningful new attack surface, so most of this is about them. **CI:** Build / Markdown / relative-link / Go race / TS unit all green; **Tauri Rust tests, TS E2E, Ollama E2E still pending** at time of review. Given this PR adds a Tauri command, the Tauri Rust job is load-bearing — please wait for it before merge.

### Security & correctness — backend + Tauri

**1. Loopback enforcement is correct but ignores `X-Forwarded-For` / `X-Real-IP`.**
`isLoopbackRemoteAddr()` validates the socket's `RemoteAddr` (IPv4 + IPv6 + malformed all covered by `TestWorkspaceOpenLoopbackDetection`). What it doesn't do is reject requests that carry an `X-Forwarded-For` or `X-Real-IP` header. If an operator ever fronts Hecate with a reverse proxy that doesn't strip those (nginx default, Caddy in some configs, ad-hoc dev tunnels), an external request landing on the loopback socket via the proxy passes the check — and then launches an editor on the operator's machine. Two minimal mitigations:
- Reject the request when `X-Forwarded-For` or `X-Real-IP` is set on this specific endpoint (and `workspace_dialog`). Cheap, defense in depth.
- Document the "do not place Hecate behind a forwarding proxy" assumption in `docs/security.md` so the trust model is explicit.

No test currently asserts header-spoofing rejection.

**2. The PR description's "constrained to the chat session's recorded workspace" claim is inaccurate for the backend.**
`validateWorkspaceOpenRequest` → `canonicalWorkspaceDialogPath` validates path-is-a-real-directory, resolves symlinks, and rejects files — but does **not** constrain to the active chat session's recorded workspace. Any directory the gateway process can stat is acceptable. The constraint is UI-level only. Probably fine for an operator console (you control the UI), but worth either correcting the PR description or adding an opt-in server-side check that the requested path equals the session's `Workspace` field. The latter would harden against a compromised-renderer scenario.

**3. Command allowlist + no shell interpretation — clean.**
Eight hardcoded targets, fixed argv, `exec.Command` and Tauri `spawn()` both bypass the shell. No injection vector.

**4. Tauri side independently validates** (`validate_workspace_open_request` in `lib.rs`). Stricter than Go (requires absolute paths). Good defense in depth. The `allow-open-workspace-target` permission is correctly scoped to desktop only.

**5. `handler_chat_files.go` exposes only Git diffs**, not file contents. `gitrunner.Restore()` constrains revert to paths already in the diff. Right scope. Minor gap: missing tests for non-git workspace / non-existent workspace / access-error paths.

**6. Dependencies (`zenity` + transitive resource-embedding bits)** are an upgrade from osascript-only to cross-platform native pickers. No risky pull-ins. ✓

**7. Concurrency** on the dialog handler uses `atomic.Bool` single-flight via CAS — correctly tested. The new `workspace_open` handler is stateless per request. ✓

### UI/UX

**8. `WorkspaceOpenMenu` localStorage persistence has no schema versioning.** Today's payload is a single string (the target id). If/when this grows to include args or per-workspace defaults, a stored value from an older client could silently mis-restore. Trivial to add a `{ v: 1, target: ... }` envelope now; expensive to migrate later.

**9. `canOpenWorkspaceFromUI()` is hardcoded to `true`.** Won't gracefully degrade if either the Tauri command or the loopback endpoint is conditionally unavailable. If the gateway disables this endpoint in some future deploy (locked-down build flag, headless mode), the menu still renders and the click fails with an error. Either check the endpoint health on mount, or add a server-reported capability flag to the bootstrap response.

**10. `ChatRightPanel` width isn't persisted.** Each chat swap resets to initial width. For "shared resizable right panel" the operator will probably expect this to remember its size — that's the whole point of resizing it once. `localStorage` keyed on `hecate.rightPanelWidth` would cover it.

**11. `ChatRightPanel` keyboard step (24px) is disconnected from mouse-drag granularity.** Pointer drag is pixel-precise; arrow keys jump in 24-pixel increments. Minor a11y quality nit — either a smaller step (4-8px) or a configurable one.

**12. `ChatWorkspaceChangesPanel`: discard flow is correctly two-step** (per-file and discard-all both gate on a `confirmRevertPath` sentinel). Good destructive-action pattern. State machine cleanup after revert is sound.

**13. Resize limit recomputes `Math.floor(window.innerWidth * 0.58)` on every drag event** without listening to `resize`. If the window is resized mid-drag the limit shifts mid-stroke. Edge case; flag-only.

**14. Markdown fix: only bold-with-code is tested.** The italic-with-code path uses the same recursive `renderInline()` call so it works, but a one-line additional fixture pinning italic+code would lock the symmetry in. Cheap.

**15. `showCapturedDiff` deduplication** in `TranscriptMessageRow` skips the captured diff when a `files_changed` activity is present. The check is string-based (`activity.type === "files_changed"`). If activity type names ever shift, this silently re-introduces double rendering. Worth a comment naming the Go source-of-truth (`internal/api/handler_chat_activities.go` style — same pattern as #229 round-3).

**16. `workspace-open.ts` test coverage misses error paths.** Both Tauri and HTTP dispatch are tested for success; neither for `403`, `500`, or Tauri-command-not-found. Add at least one error-path test per dispatch.

### Smaller things

- **Copy polish is consistent across UI strings** — no lingering "ACP adapter" / "adapter controls" in operator-facing text. Docs still say "adapter-reported" in one spot in `external-agent-adapters.md`, which is acceptable since the docs section is about adapter mechanics.
- **`Atoms.CodeBlock` diff highlighting** adds a `DiffCode()` helper with sequential prefix matching per line. Fine at chat-diff scale; if this ever renders multi-thousand-line diffs, switch to a single regex pass.
- **`ChatView` test rewrite** drops boilerplate (good) and adds workspace-dialog + right-panel-filter coverage; no test exercises the resize handle drag.

### Verdict

The headline functionality is well-shaped and the Tauri/HTTP split is correctly defended on both sides. The single thing worth doing before merge is **#1 (X-Forwarded-For rejection or documented assumption)** — that closes a real bypass path for any operator who later sticks a proxy in front. **#9 (capability detection)** and **#10 (width persistence)** are the most operator-noticeable UI follow-ups. Everything else is nits or future-pass.

PR #238, opened 2026-05-28
Improve workspace and transcript diff rendering

chicoxyzzy:
## Summary

- Add the richer Workspace changes panel with file-tree navigation, filtering, copy/discard actions, and rich per-file diffs.
- Render fenced `diff` / `patch` blocks in chat transcripts with the shared rich diff viewer instead of plain code blocks.
- Refresh the screenshot capture flow, README screenshot description, and generated UI screenshots.

## Verification

- `cd ui && bun run lint`
- `just docs-format-check`
- `cd ui && bun run test -- TranscriptMarkdown.test.tsx ChatView.test.tsx`
- `cd ui && bun run test`
- `just screenshots`


chicoxyzzy (commented):
## Review

UI-only; all required CI green (Go/Tauri correctly skipped). Builds on the rich `DiffViewer` from #236. Two inline comments below — one is a concrete rendering bug, one is a UX-regression risk.

### Strengths

- **Timer-leak hardening across three call sites** (`ChatTranscript.copyMsg`, `Atoms.CopyBtn`, `Atoms.CodeBlock`) is genuinely good. The `setTimeout(() => setState(...), N)` pattern with no cleanup is a real "setState after unmount" source; all three now track the timer in a ref and clear on unmount and before re-arming. Nice catch beyond the headline feature.
- **File tree is well-built.** `buildWorkspaceFileTree` uses a draft-node Map then materializes with single-child directory-chain collapsing, so `src/features/chats` stays one row instead of three nested folders — the right call for a narrow side panel. `sortedWorkspaceTreeEntries` (folders first, alphabetical within) and the recursive `fileCount` are clean.
- **Search filters at tree-build time** and the `useMemo` keys correctly on `[files, query]`; empty-result state is handled.
- **Context menu** closes on click/keydown via window listeners cleaned up in the effect return, and clamps to the viewport edge on open.
- **Markdown diff fences** (` ```diff ` / ` ```patch `) now render with the shared `DiffViewer` (`compact embedded`), consistent with the workspace panel. Test pins it (no `<pre>`, one `diffs-container`).
- Test coverage extended for folder collapse/expand and search filtering.

### Concerns

**1. Scope: undisclosed governance-doc changes bundled into a diff-rendering PR.** The title and summary describe only the UI/screenshot work, but the diff also adds an **RTK shell-command policy** to four governance files — `AGENTS.md`, `docs-ai/core/workflow.md`, `docs-ai/core/verification.md`, `docs/development.md`. Those additions read as reasonable, but: (a) they're invisible in the PR summary, so a description-skimming reviewer won't look at them; (b) on squash-merge the commit message comes from the PR body, so the permanent repo record won't explain why `AGENTS.md` changed; (c) #236 explicitly said "`AGENTS.md` has an unrelated local edit ... intentionally not included" — this PR now includes that previously-excluded edit silently. Agent-instruction changes deserve their own PR or at least a summary section. This is the one thing I'd ask to change before merge.

**2. Folder auto-expand resets manual collapse on every refresh/keystroke** — inline comment at the effect.

**3. Undefined CSS custom properties** for renamed/modified file colors — inline comment at `fileStatusColor`.

**4. Context menu only closes on click/keydown, not scroll/resize.** It's `position: fixed` with viewport coords captured at open time, so scrolling the panel or window while it's open leaves it pinned to a stale screen position. A capture-phase `scroll`-to-close listener would tighten it. Minor — the copy/discard actions are also available as inline row buttons, so the menu is an accelerator, not the only path.

### Verdict

Solid feature work with a real bonus fix (timer leaks). Blocking-ish item is **#1 (split or document the governance docs)**; the `var(--yellow)`/`var(--cyan)` bug (#3) is small but user-visible and worth fixing here. The auto-expand reset (#2) and context-menu scroll (#4) are follow-ups.

ui/src/features/chats/ChatWorkspaceChangesPanel.tsx
```diff
@@ -234,7 +240,23 @@ export function ChatWorkspaceChangesPanel({
   const diff = snapshot?.diff?.trim() ?? "";
   const hasChanges = Boolean(snapshot?.has_changes && (diffStat || diff));
   const summary = diffStat ? formatDiffStatSummary(diffStat) : "";
-  const files = snapshot?.files ?? [];
+  const files = snapshot?.files ?? EMPTY_CHANGED_FILES;
+  const tree = useMemo(() => buildWorkspaceFileTree(files, query), [files, query]);
+
+  useEffect(() => {
+    setExpandedDirPaths(collectWorkspaceTreeFolderPaths(tree));
```
chicoxyzzy: This force-expands **all** folders whenever `tree` changes identity — and `tree` is a `useMemo` keyed on `[files, query]`, so it recomputes on every search keystroke *and* every workspace refresh. Net effect: if the operator collapses a noisy folder and then hits Refresh (or the snapshot updates after a discard), every folder springs back open and their collapse is lost.

For the search case, expanding to reveal matches is arguably desirable. For a plain refresh it's surprising. Consider keying the auto-expand on the actual set of folder paths (so it only re-runs when the file set really changes), or merging with prior `expandedDirPaths` instead of replacing it outright.

ui/src/features/chats/ChatWorkspaceChangesPanel.tsx
```diff
@@ -457,152 +514,570 @@ function WorkspaceDiffPanel({
...
+    case "new":
+    case "untracked":
+      return "var(--green)";
+    case "deleted":
+    case "removed":
+      return "var(--red)";
+    case "renamed":
+      return "var(--yellow)";
+    default:
+      return "var(--cyan)";
```
chicoxyzzy: `--yellow` and `--cyan` are not defined anywhere in `ui/src/styles.css` — the palette only ships `--teal` / `--amber` / `--red` / `--green` (each with `-lo`/`-bg`/`-border` variants). `color: var(--yellow)` with no fallback and no definition is an invalid declaration, so the glyph falls back to the inherited color instead of a distinct status color.

That means **renamed** files and — via the `default` branch — **modified** files (the most common status) render without their intended color. Suggest `--amber` for renamed and `--teal` for modified (both defined), or add `--yellow`/`--cyan` to the palette. A two-line fix, but it's the common path so worth doing here.

ui/src/features/chats/ChatWorkspaceChangesPanel.tsx
chicoxyzzy, line 247: This force-expands **all** folders whenever `tree` changes identity — and `tree` is a `useMemo` keyed on `[files, query]`, so it recomputes on every search keystroke *and* every workspace refresh. Net effect: if the operator collapses a noisy folder and then hits Refresh (or the snapshot updates after a discard), every folder springs back open and their collapse is lost.

For the search case, expanding to reveal matches is arguably desirable. For a plain refresh it's surprising. Consider keying the auto-expand on the actual set of folder paths (so it only re-runs when the file set really changes), or merging with prior `expandedDirPaths` instead of replacing it outright.

ui/src/features/chats/ChatWorkspaceChangesPanel.tsx
chicoxyzzy, line 1077: `--yellow` and `--cyan` are not defined anywhere in `ui/src/styles.css` — the palette only ships `--teal` / `--amber` / `--red` / `--green` (each with `-lo`/`-bg`/`-border` variants). `color: var(--yellow)` with no fallback and no definition is an invalid declaration, so the glyph falls back to the inherited color instead of a distinct status color.

That means **renamed** files and — via the `default` branch — **modified** files (the most common status) render without their intended color. Suggest `--amber` for renamed and `--teal` for modified (both defined), or add `--yellow`/`--cyan` to the palette. A two-line fix, but it's the common path so worth doing here.

chicoxyzzy (commented):
## Review

UI-only; all required CI green (Go/Tauri correctly skipped). Builds on the rich `DiffViewer` from #236. Two inline comments below — one is a concrete rendering bug, one is a UX-regression risk.

### Strengths

- **Timer-leak hardening across three call sites** (`ChatTranscript.copyMsg`, `Atoms.CopyBtn`, `Atoms.CodeBlock`) is genuinely good. The `setTimeout(() => setState(...), N)` pattern with no cleanup is a real "setState after unmount" source; all three now track the timer in a ref and clear on unmount and before re-arming. Nice catch beyond the headline feature.
- **File tree is well-built.** `buildWorkspaceFileTree` uses a draft-node Map then materializes with single-child directory-chain collapsing, so `src/features/chats` stays one row instead of three nested folders — the right call for a narrow side panel. `sortedWorkspaceTreeEntries` (folders first, alphabetical within) and the recursive `fileCount` are clean.
- **Search filters at tree-build time** and the `useMemo` keys correctly on `[files, query]`; empty-result state is handled.
- **Context menu** closes on click/keydown via window listeners cleaned up in the effect return, and clamps to the viewport edge on open.
- **Markdown diff fences** (` ```diff ` / ` ```patch `) now render with the shared `DiffViewer` (`compact embedded`), consistent with the workspace panel. Test pins it (no `<pre>`, one `diffs-container`).
- Test coverage extended for folder collapse/expand and search filtering.

### Concerns

**1. Scope: undisclosed governance-doc changes bundled into a diff-rendering PR.** The title and summary describe only the UI/screenshot work, but the diff also adds an **RTK shell-command policy** to four governance files — `AGENTS.md`, `docs-ai/core/workflow.md`, `docs-ai/core/verification.md`, `docs/development.md`. Those additions read as reasonable, but: (a) they're invisible in the PR summary, so a description-skimming reviewer won't look at them; (b) on squash-merge the commit message comes from the PR body, so the permanent repo record won't explain why `AGENTS.md` changed; (c) #236 explicitly said "`AGENTS.md` has an unrelated local edit ... intentionally not included" — this PR now includes that previously-excluded edit silently. Agent-instruction changes deserve their own PR or at least a summary section. This is the one thing I'd ask to change before merge.

**2. Folder auto-expand resets manual collapse on every refresh/keystroke** — inline comment at the effect.

**3. Undefined CSS custom properties** for renamed/modified file colors — inline comment at `fileStatusColor`.

**4. Context menu only closes on click/keydown, not scroll/resize.** It's `position: fixed` with viewport coords captured at open time, so scrolling the panel or window while it's open leaves it pinned to a stale screen position. A capture-phase `scroll`-to-close listener would tighten it. Minor — the copy/discard actions are also available as inline row buttons, so the menu is an accelerator, not the only path.

### Verdict

Solid feature work with a real bonus fix (timer leaks). Blocking-ish item is **#1 (split or document the governance docs)**; the `var(--yellow)`/`var(--cyan)` bug (#3) is small but user-visible and worth fixing here. The auto-expand reset (#2) and context-menu scroll (#4) are follow-ups.

PR #245, opened 2026-05-29
refactor(api): split task run stream projection

chicoxyzzy:
## Summary

- Split task-run SSE projection out of `HandleTaskRunStream` into a focused projector that maps persisted run events plus live task storage into `TaskRunStreamEventData`.
- Split SSE frame formatting into a small writer seam, keeping the HTTP handler focused on request setup, polling, cancellation, and stream control flow.
- Keep the stream contract forward-moving: new snapshots carry the current stream shape, while older alpha rows replay as stored instead of being normalized at read time.
- Reuse the projector for Hecate Chat task activity projection and document the new seams.

## Tests

- `go test ./internal/api -run 'TestTaskRunStreamProjector|TestTaskRunStreamWriter|TestDecodeTurnCost' -count=1`
- `go test ./internal/api -count=1`
- `go test ./...`
- `go vet ./...`
- `go test -tags e2e ./e2e/... -run 'TestTaskApproval(Cancel|Reject)TerminalStateE2E' -count=1 -timeout 5m`
- `go test -tags e2e ./e2e/... -count=1 -timeout 5m`
- `go test -race -timeout 10m ./...`
- `just docs-check`
- `git diff --check`

### DIFF
--- a/docs-ai/skills/backend/SKILL.md
+++ b/docs-ai/skills/backend/SKILL.md
@@ -77,6 +77,20 @@ The seven-step chain spans `pkg/types/` → `internal/api/` → `internal/provid
 3. Update the `docs/mcp.md` tool table.
 4. Tests in `internal/mcp/server/tools_test.go` using the `fakeGateway` helper.
 
+### Change task-run streaming
+
+`GET /hecate/v1/tasks/{id}/runs/{run_id}/stream` has two seams:
+
+1. `internal/api/task_run_stream_projector.go` maps persisted run events plus
+   live task storage into `TaskRunStreamEventData`.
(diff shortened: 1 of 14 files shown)


chicoxyzzy:
Addressed the optional stale-comment nit by removing the orphaned `decodeTurnCostFromEventData` doc block from `internal/api/handler_tasks.go`.

On the approval top-up behavior: yes, the forward-only/replay-as-stored contract is intentional. Current persisted snapshots carry approvals via `liveState`; old alpha rows are not normalized in the projector, and the live/terminal projection paths still carry approvals for the operator UI.

PR #246, opened 2026-05-29
refactor(api): keep task run streams read-only

chicoxyzzy:
## Summary

- Make the per-run task SSE stream read-only by removing the synthetic `snapshot` `run_event` append from `HandleTaskRunStream`.
- Keep projected live frames on the latest persisted event cursor, so resumed streams do not fabricate a higher `sequence` without a matching durable event.
- Add API and binary e2e coverage proving resumed terminal streams emit projected `snapshot`/`done` frames without appending run events.
- Update runtime/events docs and backend agent guidance with the read-only stream contract.

Stacked on #245.

## Tests

- `go test ./internal/api -run 'TestTaskRunStreamResumeWith|TestTaskRunStreamProjector|TestTaskRunStreamWriter|TestDecodeTurnCost' -count=1`
- `go test ./internal/api -count=1`
- `go test ./...`
- `go vet ./...`
- `go test -tags e2e ./e2e/... -run 'TestTaskRunStreamResumeDoesNotAppendEventsE2E|TestTaskApproval(Cancel|Reject)TerminalStateE2E' -count=1 -timeout 5m`
- `go test -tags e2e ./e2e/... -count=1 -timeout 5m` (first run hit transient sqlite `SQLITE_BUSY` under parallel load; rerun passed)
- `go test -race -timeout 10m ./...`
- `just docs-check`
- `git diff --check feature/run-stream-projector...HEAD`

### DIFF
--- a/docs-ai/skills/backend/SKILL.md
+++ b/docs-ai/skills/backend/SKILL.md
@@ -85,11 +85,13 @@ The seven-step chain spans `pkg/types/` → `internal/api/` → `internal/provid
    live task storage into `TaskRunStreamEventData`.
 2. `internal/api/task_run_stream_writer.go` writes the SSE frames.
 
-Keep the stream contract forward-moving: new `snapshot` rows should carry the
-current `TaskRunStreamEventData` shape, and older alpha rows can replay as they
-were stored. Do not mutate historical `run_event` rows; the event log is
-append-only. Handler changes should stay focused on request setup, polling, and
-cancellation.
+Keep the stream contract forward-moving: persisted snapshot payloads should
+carry the current `TaskRunStreamEventData` shape when they are written, and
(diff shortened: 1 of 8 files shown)


chicoxyzzy:
Addressed the optional micro-nit: `taskRunStreamProjector.snapshotEventData` now returns only `(map[string]any, error)`; the live stream keeps using `taskRunStreamStateJSON` directly for change detection.

Validation:
- `go test ./internal/api -run 'TestTaskRunStreamProjector|TestTaskRunStreamResumeWith' -count=1`
- `go test -tags e2e ./e2e/... -run 'TestTaskRunStreamResumeDoesNotAppendEventsE2E|TestTaskApproval(Cancel|Reject)TerminalStateE2E' -count=1 -timeout 5m`
- `git diff --check`

PR #248, opened 2026-05-29
Harden local subprocess boundaries

chicoxyzzy:
## Summary

- sanitize stdio MCP subprocess environments so gateway/provider/telemetry secrets are not inherited implicitly
- narrow external agent adapter environments to per-adapter credential families
- reject non-loopback and forwarded-client requests for MCP probe, reset-data, and shutdown endpoints
- create/repair SQLite data directories and database files with owner-only POSIX modes
- document the updated MCP, external-agent, runtime API, and security boundaries

## Validation

- `go test ./internal/mcp/client`
- `go test ./internal/mcp/client ./internal/agentadapters ./internal/storage ./internal/api`
- `go test ./...`
- `./ui/node_modules/.bin/oxfmt --check docs/mcp.md docs/runtime-api.md docs/security.md docs/external-agent-adapters.md`

## Notes

The worktree had unrelated Tauri/UI edits before this PR; this branch only stages and commits the security hardening files.


chicoxyzzy:
Review finding: blocking until fixed.

- [P1] Do not chmod arbitrary SQLite parent directories — `internal/storage/sqlite.go:83`

  `NewSQLiteClient` now runs `os.Chmod(dir, 0o700)` for whatever parent directory appears in `HECATE_SQLITE_PATH`, even when that directory already existed and is not Hecate-owned. That means paths like `/tmp/hecate.db`, `/var/lib/shared/hecate.db`, a mounted volume root, or any operator-provided shared parent can be silently locked down by gateway startup; in a root/container deployment this could even chmod `/tmp` to `0700`. For non-owned parents it can also turn a previously working config into a startup failure. The hardening should only apply to directories Hecate creates/owns, or it should require/use a dedicated private data directory rather than chmodding an existing arbitrary parent. Please also add a regression test where an existing `0755` parent is passed and is not mutated.

Verification I ran:
- `git diff --check 249daf2d99f0dffaf9810fad30ea7f3b4db9e4e8..HEAD` — clean
- `go test ./internal/mcp/client` — 62 passed
- `go test ./internal/agentadapters ./internal/storage ./internal/api` — 743 passed
- `go test ./...` — 2038 passed
- `./ui/node_modules/.bin/oxfmt --check docs/mcp.md docs/runtime-api.md docs/security.md docs/external-agent-adapters.md` — clean

chicoxyzzy:
Addressed the env-boundary decision point:

- added non-secret operational defaults for `SSL_CERT_FILE`, `SSL_CERT_DIR`, `NODE_EXTRA_CA_CERTS`, `APPDATA`, and `LOCALAPPDATA` to MCP stdio and external-adapter runtime env allowlists
- kept `HTTP_PROXY` / `HTTPS_PROXY` / `NO_PROXY` opt-in per MCP server because proxy URLs often carry credentials
- updated `docs/mcp.md` to call out proxy settings as explicit per-server env
- clarified `docs/runtime-api.md` that Docker shutdown via published ports is rejected by the loopback guard; use in-container localhost or normal orchestrator stop paths

Validation after the update:

- `go test ./internal/mcp/client ./internal/agentadapters`
- `./ui/node_modules/.bin/oxfmt --check docs/mcp.md docs/runtime-api.md`

chicoxyzzy:
Fixed the SQLite chmod blocker:

- Hecate no longer chmods an existing parent directory for `HECATE_SQLITE_PATH` / SQLite config paths
- missing SQLite directories are still created with `0700`
- SQLite DB files are still created/repaired as `0600`
- added a regression test that verifies an existing `0755` parent directory remains unchanged
- updated `docs/security.md` to describe the more precise behavior

Validation:

- `go test ./internal/storage`
- `./ui/node_modules/.bin/oxfmt --check docs/security.md`

PR #262, opened 2026-05-30
refactor(providers): drop static seed-list from BuiltInProvider

chicoxyzzy:
## Summary

Every cloud preset shipped a curated \`Models: []string{...}\` list that was copied into \`OpenAICompatibleProviderConfig.KnownModels\` at runtime and surfaced via \`staticCapabilities()\` as a fallback when no API key was set or live discovery was skipped.

The intent was a nice cold-start UX (\"see what models you'll get before you configure\"). In practice the lists bit-rot the moment upstream catalogs churn:

- Groq listed \`mixtral-8x7b-32768\` — **deprecated by Groq**.
- Anthropic listed \`claude-opus-4-7\` — **does not exist**.
- Together / Fireworks-class providers churn weekly; any seed list lies within days of shipping.

The runtime already prefers \`/v1/models\` discovery over the seed list, and the API surface (\`/hecate/v1/providers/presets\`) does **not** expose Models to clients — so removing the field has zero external contract impact. It just stops carrying around a maintenance burden whose only job was to look pretty before the operator finished onboarding.

## Diff

```
internal/config/config.go             |  7 +++--
internal/config/provider_catalog.go   | 49 -----------------------------------
internal/providers/openai.go          |  7 +++--
internal/providers/runtime_manager.go |  1 -
4 files changed, 10 insertions(+), 54 deletions(-)
```

## What changes

- **\`BuiltInProvider.Models\` field gone** — and all nine \`Models: []string{...}\` entries with it (anthropic, deepseek, gemini, groq, mistral, openai, perplexity, together_ai, xai). The four local presets and Fireworks already had no seed list — they now match the pattern.
- **\`RuntimeConfig()\` no longer copies** \`KnownModels\` from the catalog.
- **\`runtime_manager.go\` no longer mirrors** the catalog seed into per-runtime config.
- **\`KnownModels\` field stays on \`OpenAICompatibleProviderConfig\`** — \`PROVIDER_<NAME>_MODELS\` env still populates it as an operator escape hatch (useful for air-gapped deployments where discovery is unavailable).
- **\`staticCapabilities\` comment updated** in \`openai.go\` to reflect the new world.

## User-visible impact on cold start

| Operator state | Before | After |
|---|---|---|
| Added cloud preset, no API key | Picker shows curated list (often lying about deprecated/missing models) | Picker empty until they paste the key |
| Added cloud preset + valid API key | Discovery runs, real catalog populates | **Unchanged** |
| Set \`PROVIDER_OPENAI_MODELS=gpt-4o,gpt-4.1\` env | Override wins | **Unchanged** |

## Verification

- \`go build ./...\` — succeeds.
- \`go test ./internal/config/... ./internal/providers/... ./internal/api/...\` — pass.
- \`go test -race ./...\` — all packages pass.

## Test plan

- [ ] CI green (\`test.yml\` runs the full Go matrix).
- [ ] Smoke: add OpenAI preset without API key → confirm empty picker (the new behavior).
- [ ] Smoke: add OpenAI preset with API key → confirm discovery populates real models.
- [ ] Smoke: set \`PROVIDER_OPENAI_MODELS=test-1,test-2\` → confirm override still works.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -323,8 +323,11 @@ type OpenAICompatibleProviderConfig struct {
 	StubResponse string        `json:"stub_response"`
 	DefaultModel string        `json:"default_model"`
 	Enabled      bool          `json:"enabled"`
-	// KnownModels is the curated catalog from the built-in preset. It populates the
-	// static capabilities when no API key is set and live discovery is skipped.
+	// KnownModels is an operator-supplied static catalog override. Populated only
+	// via PROVIDER_<NAME>_MODELS env (comma-separated), it populates the static
+	// capabilities when no API key is set and live discovery is skipped. Empty
+	// by default — Hecate prefers live `/v1/models` discovery over hard-coded
+	// lists that bit-rot as upstream catalogs churn.
 	KnownModels []string `json:"known_models,omitempty"`
 	// AnthropicCacheDisabled opts the Anthropic adapter out of
 	// auto-attaching `cache_control: {"type":"ephemeral"}` markers on
--- a/internal/config/provider_catalog.go
+++ b/internal/config/provider_catalog.go
@@ -21,12 +21,6 @@ type BuiltInProvider struct {
 	DocsURL      string
 	Description  string
 	StubResponse string
-	// Models is a curated list of well-known model IDs for the provider, used as
-	// the static catalog when API-driven discovery isn't possible (no API key, or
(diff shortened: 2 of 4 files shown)


chicoxyzzy:
Superseded by a larger BuiltInProvider cleanup that also drops DefaultModel and StubResponse and improves descriptions. New PR incoming.

PR #263, opened 2026-05-30
refactor(providers): drop per-preset StubResponse from BuiltInProvider

chicoxyzzy:
## Summary

Every \`BuiltInProvider\` entry hardcoded a \`StubResponse\` string that was copied into \`OpenAICompatibleProviderConfig.StubResponse\` via \`RuntimeConfig()\`. 14 entries with only **three distinct strings**:

- **9 cloud presets**: \`\"Stubbed response from the AI Agent Runtime MVP.\"\` — identical to the default that \`normalizeProviders\` already applies in \`config.go:952-953\`.
- **4 local presets**: \`\"Stubbed local provider response.\"\`
- **1 anthropic preset**: \`\"Stubbed Anthropic response.\"\`

The per-preset override only takes effect for env-driven providers (\`PROVIDER_<NAME>_STUB_MODE=true\` + matching \`_STUB_RESPONSE\`). Operators rarely flip stub mode, and when they do the prefix string is near-cosmetic — the echo content is the actual signal.

## What changes

| | Before | After |
|---|---|---|
| \`BuiltInProvider.StubResponse\` field | exists | removed |
| 14 catalog \`StubResponse:\` entries | populated | removed |
| \`RuntimeConfig()\` copy | \`StubResponse: p.StubResponse\` | dropped |
| \`OpenAICompatibleProviderConfig.StubResponse\` field | exists | **unchanged** — still env-configurable via \`PROVIDER_<NAME>_STUB_RESPONSE\` |
| \`OpenAICompatibleProviderConfig.StubMode\` field | exists | **unchanged** |
| \`normalizeProviders\` default-fill | supplies the generic string | **unchanged** — still the single source of truth |

## Behavior

For the rare operator running \`PROVIDER_ANTHROPIC_STUB_MODE=true\`:

- **Before:** echoes start with \`\"Stubbed Anthropic response. Echo: ...\"\`.
- **After:** echoes start with \`\"Stubbed response from the AI Agent Runtime MVP. Echo: ...\"\` (the generic default-fill string).

Same applies to the four local presets. The bespoke prefixes are gone; the response shape is unchanged.

## Diff stat

```
1 file changed, 104 insertions(+), 119 deletions(-)
```

The large insertion count is gofmt re-aligning struct field colons after the longest field name (\`StubResponse\`) was removed — net effect is 15 logical lines removed.

## Relationship to PR #262

Independent. #262 removes \`Models []string\`; this removes \`StubResponse string\`. Either can land first; whoever's second rebases trivially.

## Verification

- \`go build ./...\` — clean.
- \`go test ./internal/config/... ./internal/providers/... ./internal/api/...\` — pass.
- \`go test -race ./...\` — all packages pass.


chicoxyzzy:
Superseded by a larger BuiltInProvider cleanup that also drops Models and DefaultModel and improves descriptions. New PR incoming.

PR #282, opened 2026-05-31
refactor(chat): retire ExecutionModeDirectModel as a write-side value

chicoxyzzy:
## Summary

Every Hecate-side chat message now persists as `execution_mode = "hecate_task"` regardless of tools state. The tools-on/off discriminant lives entirely on the per-message `ToolsEnabled` boolean. The legacy `"direct_model"` literal stays **accepted as a wire-side back-compat input** but is no longer written anywhere — and a sqlite migration sweeps existing rows forward so the value also disappears from new reads.

Stacked on PR #281 (the dead-flip cleanup). Net behavioral change: persisted messages from tools-off turns now carry `execution_mode = "hecate_task"` + `tools_enabled = false` instead of `execution_mode = "direct_model"`.

## Backend writes

- `internal/chat/store.go`: drop the `ExecutionModeDirectModel` constant; introduce `LegacyExecutionModeDirectModel` whose name signals to readers that the value is only here for older clients and older persisted rows.
- `defaultMessageExecutionMode` returns `hecate_task` for any Hecate-side session.
- `handleDirectModelTurn` writes `ExecutionMode: hecate_task` (with `ToolsEnabled: false`) for tools-off turns.
- `directModelContextPacket` and `defaultMessageExecutionModeForRender` return `hecate_task`.
- `normalizeChatExecutionMode` folds the legacy `direct_model` literal forward to `hecate_task` before the dispatcher's switch sees it, so the switch only has to know about the two surviving execution modes.
- `chatRequestToolsEnabled` rewritten as an explicit switch over the execution_mode literals. The pre-tools_enabled "no signal" contract (absence of both fields defaulted to model-chat dispatch) is preserved by returning `false` when both fields are empty.

## Sqlite migration

- `internal/chat/sqlite.go` gains a second backfill UPDATE that rewrites every `execution_mode = 'direct_model'` row to `'hecate_task'`. Runs **after** the existing `tools_enabled = 0` backfill so the tools-off intent is preserved on every row before the discriminant changes shape. Idempotent.
- `internal/chat/sqlite_test.go` extends the existing backfill conformance to also assert the execution_mode flip.

## Segment continuation

- `shouldStartNewHecateAgentSegment` checked `message.ExecutionMode != chat.ExecutionModeHecateTask` to detect a break in the running tool-task segment. After the rename every Hecate-side message matches that, so the function would refuse to start a new task across tools-off → tools-on transitions. Replace the check with `isHecateTaskToolsOnMessage`, which inspects both `ExecutionMode` and `ToolsEnabled` and recognizes the legacy literal too.
- `modelSegmentID` gets the same treatment via `messageIsDirectModelTurn`.

## Dispatcher

The switch case drops `chat.ExecutionModeDirectModel` because the legacy literal is normalized away before the switch sees it. The dispatcher only knows `hecate_task` and `external_agent` now.

## Tests

- `handler_chat_dispatch_test.go` rewritten to assert the new semantics: explicit mode wins, legacy `direct_model` normalizes forward, external sessions pin `external_agent`, Hecate sessions default to `hecate_task`. Plus a dedicated table-driven test for `chatRequestToolsEnabled` exercising the explicit-true / explicit-false / legacy-direct_model / hecate_task / both-empty matrix.
- `handler_chat_hecate_test.go` updates 8 sites that previously asserted `ExecutionMode == direct_model` on persisted messages.

## UI

**No UI changes.** `ChatExecutionMode` already accepts `"direct_model"` as a legacy literal, the UI doesn't derive toggle state from message history anyway (`useChatToolsEnabled` reads slice state, not the wire), and reads of pre-migration data still display correctly. The execution_mode axis on the wire collapses to two values for new writes; clients tolerate the third forever.

## Out of scope (still owed)

- **`LegacyExecutionModeDirectModel` constant stays for one cycle.** Once every reachable install has restarted (sweeping the sqlite backfill) and every client has stopped sending the literal, the constant + its three read-side guards can be deleted in a tiny follow-up.
- **Docs sweep.** `docs/chat-sessions.md` + `docs/runtime-api.md` still describe `direct_model` as a valid execution_mode value. Accurate as wire input (back-compat), but worth a sweep that surfaces "tools-off Hecate-task turns" as the new canonical encoding.

## Verification

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test -race -timeout 5m ./internal/api/... ./internal/chat/...` — all green
- [x] `bun run typecheck` (UI) — clean
- [x] `bun run test` (UI) — 1163 / 1163 pass
- [x] `bunx playwright test chat.spec.ts` — 44 / 44 pass
- [x] `bun run lint` (UI) — 0 warnings / 0 errors
- [x] `bun run format:check` (UI) — clean

## Diff stat

```
9 files changed, 253 insertions(+), 120 deletions(-)
```

### DIFF
--- a/internal/api/handler_chat.go
+++ b/internal/api/handler_chat.go
@@ -751,25 +751,31 @@ func (h *Handler) HandleCreateChatMessage(w http.ResponseWriter, r *http.Request
 		return
 	}
 	executionMode := normalizeChatExecutionMode(req.ExecutionMode, session, req.ToolsEnabled)
-	// Resolve the effective tools-on/off signal for this turn. The
-	// capability-driven downgrade (a tool-incapable model on a Hecate
-	// session) flips this to false even when the client requested
-	// tools-on, so the unified handler below routes the turn to the
-	// direct-model code path without the dispatcher needing a
-	// separate direct_model switch case.
-	toolsEnabled := executionMode == chat.ExecutionModeHecateTask
+	// Resolve the effective tools-on/off signal for this turn from
+	// either the new `tools_enabled` field or the legacy
+	// `execution_mode = "direct_model"` literal. The capability-
+	// driven downgrade (a tool-incapable model on a Hecate session)
+	// flips this to false even when the client requested tools-on,
+	// so the unified handler routes the turn to the direct-model
+	// code path without the dispatcher needing a separate
+	// direct_model switch case.
+	toolsEnabled := chatRequestToolsEnabled(req)
 	if toolsEnabled && !isExternalChatSession(session) && h.hecateTaskShouldFallbackToDirectModel(r.Context(), session, req) {
 		toolsEnabled = false
 	}
 	switch executionMode {
-	case chat.ExecutionModeDirectModel, chat.ExecutionModeHecateTask:
+	case chat.ExecutionModeHecateTask:
 		// One unified entry point for every Hecate-side turn,
 		// regardless of tools_enabled. handleCreateHecateChatMessage
 		// branches at the top: tools-off delegates to
 		// `handleDirectModelTurn` (the former
 		// handleCreateModelChatMessage, now a private sub-path of
 		// the Hecate handler); tools-on runs the existing
-		// agent_loop task-creation path.
+		// agent_loop task-creation path. The legacy
+		// `execution_mode = "direct_model"` literal that older
+		// clients still send normalizes to `hecate_task` inside
+		// `normalizeChatExecutionMode`, so it falls into this case
+		// the same way an explicit hecate_task would.
 		if isExternalChatSession(session) {
 			writeAgentChatRuntimeMismatch(w, "external agent sessions cannot run Hecate Chat turns")
 			return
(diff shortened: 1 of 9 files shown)


chicoxyzzy:
Reviewed #282. No blocking findings.

I checked the execution-mode retirement path end to end: legacy `execution_mode: "direct_model"` still preserves tools-off intent via `chatRequestToolsEnabled`, normalizes to Hecate-task dispatch, new tools-off writes persist as `execution_mode: "hecate_task"` + `tools_enabled: false`, segment continuation uses the boolean rather than the retired literal, and the SQLite migration backfills `tools_enabled` before rewriting legacy rows.

Verification I reran:
- `git diff --check $(git merge-base 66272e1c10823e1816444842aeecd95358f6db30 HEAD)..HEAD` — clean
- `go build ./...` — clean
- `go vet ./...` — clean
- `go test -race -timeout 5m ./internal/api/... ./internal/chat/...` — 533 passed
- `go test ./...` — 2148 passed
- `cd ui && bun run test` — 52 files / 1163 tests passed
- `cd ui && bun run lint` — 0 warnings / 0 errors
- `cd ui && bun run format:check` — clean
- `cd ui && bunx playwright test chat.spec.ts` — 44/44 passed

PR #283, opened 2026-05-31
docs(chat): surface tools_enabled as the canonical tools-on/off axis

chicoxyzzy:
## Summary

The model-chat removal sequence (PR #265 → #282) ended with every Hecate-side message persisting as `execution_mode="hecate_task"` and the tools-on/off intent on a separate per-message `tools_enabled` boolean. The operator docs were still describing `execution_mode="direct_model"` as a canonical encoding rather than a back-compat input. This PR updates them.

**Stacked on #282** — depends on the write-side retirement landing first so the docs accurately describe what the gateway actually does.

### `docs/runtime-api.md`

- The `agent_id="hecate"` description now says turns set `execution_mode="hecate_task"` + `tools_enabled`, with the legacy `direct_model` literal explicitly framed as "still accepted as a back-compat input."
- `POST /hecate/v1/chat/sessions/{id}/messages` documents `tools_enabled` as a first-class per-turn override, with explicit derivation rules for clients that omit it (legacy `direct_model` → `false`; `hecate_task` → `true`; both empty → `false`, matching the pre-`tools_enabled` default).
- The busy-session rejection is now worded in terms of tools-off turns rather than `direct_model`.
- The `chat.execution_mode_invalid` error row notes the legacy literal is also accepted and normalized to `hecate_task`.

### `docs/chat-sessions.md`

- "Model segments" rewritten to frame `tools_enabled=false` as the canonical encoding and `direct_model` as a wire-side back-compat input the gateway folds forward.

## Untouched

- Two RFC files under `docs/rfcs/` that mention the old `ChatExecutionMode` union. They're historical record of when the decision landed and shouldn't be edited post-facto.
- `docs/telemetry.md` doesn't reference `direct_model` anywhere — no change needed.

## Verification

- [x] `bun scripts/check-mermaid.ts` — clean
- [x] `oxfmt --check` against all tracked markdown — clean

## Diff stat

```
2 files changed, 56 insertions(+), 35 deletions(-)
```

## What this finishes

The "docs sweep" follow-up tracked in #282's PR description. After this lands, only one follow-up remains in the model-chat removal sequence: **delete `LegacyExecutionModeDirectModel`** once every reachable install has restarted (sweeping the sqlite backfill) and every client has stopped sending the literal. That's a tiny post-cycle cleanup, not load-bearing for any near-term work.


chicoxyzzy:
Reviewed #283. No blocking findings.

The updated docs line up with #282’s new shape: Hecate-side persisted messages are described as `execution_mode="hecate_task"` with `tools_enabled` carrying tools-on/off intent, and legacy `direct_model` is framed as a back-compat input rather than canonical storage.

Small non-blocking cleanup: `docs/runtime-api.md` still has one error-table row for `chat.agent_session_busy` that says "even for direct model turns in the same Hecate Chat session." Since this PR is moving the canonical wording to `tools_enabled=false`, that row should probably mirror the new language (for example, "even for tools-off turns").

Verification I ran:
- `git diff --check $(git merge-base aa109ef2968945ca23c362ad51e8cc4ab5be7c71 HEAD)..HEAD` — clean
- `bun scripts/check-mermaid.ts` — 69 markdown files checked, no Mermaid issues
- `git ls-files "*.md" | xargs ./ui/node_modules/.bin/oxfmt --check` — clean

PR #306, opened 2026-06-04
Add agent handoffs

chicoxyzzy:
## Summary

Adds a conservative V1 structured handoff primitive for project work so one assignment/run/chat can pass context and a recommended next action to another role or assignment while keeping follow-up execution operator-controlled.

## Changes

- Add project-scoped handoff records with memory and SQLite store parity.
- Add Hecate-native handoff list/create/update/status/delete API endpoints with project activity handoff projections.
- Add Projects UI handoff management plus explicit target-assignment and start-from-handoff actions.
- Update runtime API docs, event catalog notes, and project/memory/context RFCs for the V1 boundaries.

## Validation

- go test ./internal/projectwork ./internal/api
- go vet ./internal/projectwork ./internal/api
- go test -race -timeout 10m ./...
- cd ui && bun run typecheck
- cd ui && bun run lint
- cd ui && bun run format:check
- cd ui && bun run test
- git diff --check

### DIFF
--- a/internal/api/handler_project_work.go
+++ b/internal/api/handler_project_work.go
@@ -2017,11 +2017,18 @@ func groupProjectActivityHandoffs(handoffs []projectwork.Handoff) (map[string][]
 	byAssignment := make(map[string][]projectwork.Handoff)
 	byWorkItem := make(map[string][]projectwork.Handoff)
 	for _, handoff := range handoffs {
+		attached := false
 		if handoff.SourceAssignmentID != "" {
 			byAssignment[handoff.SourceAssignmentID] = append(byAssignment[handoff.SourceAssignmentID], handoff)
-			continue
+			attached = true
 		}
-		byWorkItem[handoff.WorkItemID] = append(byWorkItem[handoff.WorkItemID], handoff)
+		if handoff.TargetAssignmentID != "" && handoff.TargetAssignmentID != handoff.SourceAssignmentID {
+			byAssignment[handoff.TargetAssignmentID] = append(byAssignment[handoff.TargetAssignmentID], handoff)
+			attached = true
+		}
+		if !attached {
+			byWorkItem[handoff.WorkItemID] = append(byWorkItem[handoff.WorkItemID], handoff)
+		}
 	}
 	return byAssignment, byWorkItem
 }
--- a/internal/api/handler_project_work.go
+++ b/internal/api/handler_project_work.go
@@ -99,6 +99,51 @@ type createProjectWorkArtifactRequest struct {
 	AuthorRoleID string `json:"author_role_id,omitempty"`
 }
 
+type createProjectHandoffRequest struct {
+	ID                    string   `json:"id,omitempty"`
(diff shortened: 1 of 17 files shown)


chicoxyzzy (commented):
Found two issues that should be fixed before merge. Validation run on head `7cbaa9f`: `go test ./internal/projectwork ./internal/api` passed outside the sandbox after the sandbox blocked `httptest` port binding; `go vet ./internal/projectwork ./internal/api`, `git diff --check`, `cd ui && bun run typecheck`, `cd ui && bun run lint`, `cd ui && bun run format:check`, and `cd ui && bun run test` all passed.

internal/api/handler_project_work.go
```diff
@@ -1671,6 +1969,63 @@ func groupProjectActivityArtifacts(artifacts []projectwork.CollaborationArtifact
...
+		recent = append(recent, renderProjectHandoff(handoff))
+	}
+	return summary, recent
+}
+
+func groupProjectActivityHandoffs(handoffs []projectwork.Handoff) (map[string][]projectwork.Handoff, map[string][]projectwork.Handoff) {
+	byAssignment := make(map[string][]projectwork.Handoff)
+	byWorkItem := make(map[string][]projectwork.Handoff)
+	for _, handoff := range handoffs {
+		if handoff.SourceAssignmentID != "" {
```
chicoxyzzy, line 2021: [P2] Please index handoffs by `TargetAssignmentID` as well as `SourceAssignmentID`. `renderProjectActivity` only reads `handoffsByAssignment[projected.ID]`, but this grouping puts a handoff in that map only for the source assignment and then `continue`s. So a handoff targeting `asgn_review` either disappears from the review assignment’s activity row entirely, or shows only on the source row when both source and target are set, despite the API docs promising projections for the same source or target assignment. The follow-up assignment is exactly where operators need the handoff signal.

ui/src/features/projects/ProjectsView.tsx
```diff
@@ -739,6 +772,122 @@ export function ProjectsView({ onOpenChat, onOpenTask }: Props) {
...
+  }
+
+  async function handleCreateAssignmentFromHandoff(handoff: ProjectHandoffRecord) {
+    if (!selectedProjectID || !selectedWorkItemID) return;
+    const roleID = (handoff.target_role_id || "software_developer").trim();
+    if (!roleID) return;
+    setHandoffActionID(handoff.id);
+    setHandoffError("");
+    try {
+      const assignment = await createProjectAssignment(selectedProjectID, selectedWorkItemID, {
```
chicoxyzzy: [P2] This action ignores `handoff.target_work_item_id` and always creates the target assignment on the currently selected/source work item. An API-created handoff can carry `target_work_item_id` for a follow-up work item; clicking `Target assignment` would create the assignment under the wrong work item, then mark this handoff accepted with that mismatched assignment id. Please create against `handoff.target_work_item_id || selectedWorkItemID` and refresh the affected detail, or disable this action for cross-work-item handoffs.

PR #307, opened 2026-06-04
feat(projects): add memory promotion flow

chicoxyzzy:
## Summary

Adds an operator-controlled memory promotion flow for project memory:

- Adds project-scoped memory candidates with memory and SQLite storage parity.
- Adds candidate API endpoints for list/create/promote/reject, where only explicit promotion creates durable project memory.
- Adds Projects UI review controls to inspect candidates, edit before promotion, and reject/dismiss.
- Updates runtime API and memory/context RFC docs to document the promotion boundary.

## Verification

- go build ./...
- go test ./internal/memory ./internal/api
- go vet ./internal/memory ./internal/api
- go test -race -timeout 10m ./internal/memory ./internal/api
- cd ui && bun run typecheck
- cd ui && bun run test -- src/lib/api.test.ts src/features/projects/ProjectsView.test.tsx
- cd ui && bun run lint
- cd ui && bun run format:check
- git diff --check


internal/api/handler_project_memory.go
```diff
@@ -87,6 +117,184 @@ func (h *Handler) HandleCreateProjectMemoryEntry(w http.ResponseWriter, r *http.
...
+	if req.TrustLabel != nil {
+		entry.TrustLabel = *req.TrustLabel
+	}
+	if req.SourceKind != nil {
+		entry.SourceKind = *req.SourceKind
+	}
+	if req.SourceID != nil {
+		entry.SourceID = *req.SourceID
+	}
+	created, err := h.memory.Create(r.Context(), entry)
```
chicoxyzzy: [P2] Promotion needs to be atomic with marking the candidate resolved. Right now the durable memory entry is inserted here, then the candidate is updated separately below. If the candidate update fails after this succeeds, the endpoint returns 500 but leaves a new enabled memory entry behind while the candidate remains pending; a retry can create a second durable entry for the same candidate. Please move promotion into a store-level transaction/operation that creates the entry and flips the candidate together, or otherwise roll back/delete the entry when the candidate update fails.

internal/api/handler.go
```diff
@@ -390,6 +393,9 @@ func (h *Handler) SetMemoryStore(store memory.Store) {
 		return
 	}
 	h.memory = store
+	if candidates, ok := store.(memory.CandidateStore); ok {
```
chicoxyzzy, line 396: [P2] `SetMemoryStore` should clear or replace `memoryCandidates` when the injected store does not implement `CandidateStore`. As written, the handler keeps the previous candidate store, usually the default in-memory store, so memory entries and memory candidates can silently use different backends after a test/dev injection. Candidate endpoints also keep passing `requireProjectExists` even though the configured memory store cannot persist candidates. Please set `h.memoryCandidates = nil` in the non-implementing case, or require/set a combined store explicitly.

chicoxyzzy (commented):
Found two consistency issues that should be fixed before merge. Validation run on head `2a7e577`: `go build ./...`, `go vet ./internal/memory ./internal/api`, `go test ./internal/memory ./internal/api`, `go test -race -timeout 10m ./internal/memory ./internal/api`, `cd ui && bun run typecheck`, focused UI tests, lint, and format check all passed.

PR #873, opened 2026-07-17
chore(deps): pin stable ACP adapter releases

chicoxyzzy:
## Summary

- Replace development pseudo-versions with the stable v0.2.0 shared kit and built-in adapter module tags.
- Document the stable embedded-library, attachment, process-containment, and session-recovery boundaries.

## Why

The stable tags are now published and verified. Hecate already exercised equivalent production source through the previous pseudo-version pins; this change gives web and desktop builds stable, reproducible module provenance.

## Verification

- `go mod verify`
- `go test ./...`
- `go vet ./...`
- `go test -race -timeout 10m ./internal/agentadapters`
- `just docs-check`
- `just test-acp-real-embedded` against both authenticated provider CLIs, including attached-file delivery
- All twelve standalone release archives passed checksum and GitHub artifact-attestation verification

## Security and privacy

- No production adapter source delta from the previously pinned development revisions.
- Existing private staging, bounded disclosure, environment sanitization, and provider-process containment remain unchanged.


chicoxyzzy:
Closing as obsolete: current master already carries newer stable ACP adapter pins (acp-adapter-kit/Claude v0.3.1 and Codex v0.2.1). Rebasing this PR would either become empty or downgrade those versions back to v0.2.0.

